PR #37, opened 2024-12-13
feat(package): add sharepoint drive getItemsByPath and getDriveItems

AlexMoulin:
## Type of change 🧩

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Other (please explain in the description section above)

## Checklist before requesting a review

- [x] I have performed a self-review of my own code
- [x] I have double checked this work does not include any hardcoded secrets or passwords
- [x] I have made corresponding changes to the documentation
- [x] I have provided details on how I have tested my code
- [x] I have referenced the ticket number above
- [x] I have provided a description of the ticket
- [x] I have included unit tests to cover any testable code changes

PR #49, opened 2025-01-15
feat(package): add url-builder, sharepoint drive file fetching methods

AlexMoulin:
## Describe your changes
Allow for multiple, single or no query params, test response data,


beejjacobs (changes requested):
Looks like most of this is actually on main already! Might not be needed?

See #37

PR #59, opened 2025-01-20
feat(portal): Application information page

tomxelliott:
## Describe your changes

- add content for application information page
- add router handler for application information page (/application-information)

https://pins-ds.atlassian.net/browse/CROWN-8

Screenshot 2025-01-27 at 09 53 06
Screenshot 2025-01-27 at 09 53 13
Screenshot 2025-01-27 at 09 53 25

### DIFF
--- a/apps/portal/src/app/views/application-info/controller.js
+++ b/apps/portal/src/app/views/application-info/controller.js
@@ -1,4 +1,4 @@
-import { crownDevelopmentToViewModel } from 'crowndev-manage/src/app/views/cases/view/view-model.js';
+import { formatDateForDisplay } from '@pins/dynamic-forms/src/lib/date-utils.js';
 
 /**
  * @param {Object} opts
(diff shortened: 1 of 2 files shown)


beejjacobs (changes requested):
As discussed in chat, we're after the values from the DB rather than more of a guidance page. We'll likely need some content like this though.

apps/portal/src/util/config-middleware.js
```diff
@@ -16,7 +16,7 @@ export function addLocalsConfiguration(req, res, next) {
 		},
 		{
 			text: 'More Information',
-			href: '/more-infromation'
+			href: '/more-information'
```
beejjacobs, line 19: Oops, good spot!

apps/portal/src/app/views/application-info/controller.js
```diff
@@ -0,0 +1,65 @@
...
+				displayName: true
+			}
+		});
+
+		const lpa = await db.lpa.findUnique({
+			where: { id: crownDevelopment.lpaId },
+			select: {
+				name: true
+			}
+		});
```
beejjacobs, line 36: With the `include` statement you have above, this does a JOIN automatically to get the related entities, so no separate db calls required. If you add `Type: true`, you'll get the type relation, then

e.g. `crownDevelopment.Lpa` would be the LPA details, `crownDevelopment.Type.displayName` would be the type.

apps/portal/src/app/views/application-info/controller.js
```diff
@@ -0,0 +1,65 @@
...
+		.filter(([, value]) => value !== null)
+		.map(([, value]) => value)
+		.join(', ');
+}
+
+function formatDate(dateToFormat) {
+	if (!dateToFormat) {
+		return 'To be determined';
+	}
+	return new Date(dateToFormat).toDateString();
```
beejjacobs, line 64: This isn't timezone safe, FYI. You can see in the dynamic-forms solution some date utils using date-fns and date-fns-tz

apps/portal/src/app/router.js
```diff
@@ -23,6 +24,7 @@ export function buildRouter({ logger, config, dbClient }) {
 
 	router.route('/').get(viewHomepage);
 	router.use('/', projectRoutes({ db: dbClient, logger }));
+	router.use('/', createApplicationInfoRoutes({ db: dbClient, logger }));
```
beejjacobs, line 27: I'll update the other PR on this - I think we'll update the structure so its:

`views/applications`

then we can have a `list` and `view` folder in there.

apps/portal/src/app/views/application-info/controller.js
```diff
@@ -0,0 +1,65 @@
+import { crownDevelopmentToViewModel } from 'crowndev-manage/src/app/views/cases/view/view-model.js';
```
beejjacobs, line 1: We shouldn't reference the manage project like this, ideally they wouldn't depend on each other - any common code would be in packages instead.

PR #68, opened 2025-01-23
feat(crown-68)

tomxelliott:
## Describe your changes
Set up reusable gov notify client

https://pins-ds.atlassian.net.mcas.ms/browse/CROWN-68

### DIFF
--- a/packages/lib/govnotify/gov-notify-client.js
+++ b/packages/lib/govnotify/gov-notify-client.js
@@ -2,8 +2,8 @@ import { NotifyClient } from 'notifications-node-client';
 
 export class GovNotifyClient {
 	/**
-     @param {import('pino').Logger} logger
-     @param {string} [govNotifyApiKey] - Gov Notify API key
+	 * @param {import('pino').Logger} logger
+	 * @param {string} govNotifyApiKey - Gov Notify API key
 	 **/
 	constructor(logger, govNotifyApiKey) {
 		this.logger = logger;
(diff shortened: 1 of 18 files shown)


packages/lib/govnotify/gov-notify-client.js
```diff
@@ -0,0 +1,26 @@
+import { NotifyClient } from 'notifications-node-client';
+
+export class GovNotifyClient {
+	/**
+     @param {import('pino').Logger} logger
+     @param {string} [govNotifyApiKey] - Gov Notify API key
```
beejjacobs, line 6: I think your jsdoc isn't quite right somehow, with spacing and missing `*` prefix? most IDEs should generate this.

Also is the API key really optional?

packages/lib/govnotify/gov-notify-client.js
```diff
@@ -0,0 +1,26 @@
...
+     @param {string} emailAddress - Recipients email address
+     @param {import('./types.js').GovNotifyOptions} options - Options to pass to Gov Notify
+	 **/
+	async sendEmail(templateId, emailAddress, options) {
+		try {
+			this.logger.info(`dispatching email template: ${templateId}`);
+			await this.notifyClient.sendEmail(templateId, emailAddress, options);
+		} catch (e) {
+			this.logger.error(e.message);
+		}
```
beejjacobs, line 24: I think we should pass messages to the caller, else we don't know when emails fail.

packages/lib/govnotify/templates.js
```diff
@@ -0,0 +1 @@
+export const TEST_TEMPLATE = '4b8adfb1-1b7c-4333-b512-761eeedfdca2';
```
beejjacobs, line 1: Template IDs may vary by environment, so these should be configured via Terraform/infrastructure and passed in (as with the API key). e.g. `GOV_NOTIFY_TEMPLATE_X_ID`

beejjacobs (commented):
LGTM!

beejjacobs (approved):

PR #77, opened 2025-01-28
feat(packages): add sharepoint methods for adding permissions to files and folders

AlexMoulin:

### DIFF
--- a/packages/sharepoint/src/sharepoint/drives/drives.js
+++ b/packages/sharepoint/src/sharepoint/drives/drives.js
@@ -1,8 +1,8 @@
 import { UrlBuilder } from '../../util/url-builder/url-builder.js';
 
 /** @typedef {import('../../fixtures/sharepoint.js').getDriveItemsByPathData} DriveItemByPathResponse */
-/** @typedef {import('../../fixtures/sharepoint.js').listItemPermissions} ListItemPermissionsResponse */
-
+/** @typedef {Array<import('../../fixtures/sharepoint.js').ListItemPermission>} ListItemPermissionsResponse */
+/** @typedef { 'write' | 'read' | 'owner' } Role */
 /** @typedef {import('../../fixtures/sharepoint.js').getDriveItems} DriveItems */
 
 /** @typedef {Object} CopyDriveInstructions
@@ -113,11 +113,12 @@ export class SharePointDrive {
 	/**
 	 *
 	 * @param {string} itemId Id of the folder to share
-	 * @param {boolean} [requireSignIn] Specifies whether the recipient of the invitation is required to sign in to view the shared item. Defaults to true
-	 * @param {boolean} [sendInvitation] 	If true, a sharing link is sent to the recipient. Otherwise, a permission is granted directly without sending a notification. Defaults to false
-	 * @param { 'write' | 'read' } role
-	 * @param { Array<{ id: string, email: string }>} users
-	 * @param {string} [message] Message that accompanies the Invitation if sent (defaults to '')
+	 * @param {Object} options
+	 * @param {boolean} [options.requireSignIn] Specifies whether the recipient of the invitation is required to sign in to view the shared item. Defaults to true
+	 * @param {boolean} [options.sendInvitation] 	If true, a sharing link is sent to the recipient. Otherwise, a permission is granted directly without sending a notification. Defaults to false
+	 * @param { Role } options.role
+	 * @param { Array<{ id: string, email: string }>} options.users
+	 * @param {string} [options.message] Message that accompanies the Invitation if sent (defaults to '')
 	 * @returns {Promise<void>}
 	 */
 	async addItemPermissions(itemId, { requireSignIn = true, sendInvitation = false, role, users, message = '' }) {
@@ -157,7 +158,7 @@ export class SharePointDrive {
 	 *
 	 * @param {string} itemId
 	 * @param {string} permissionIdToUpdate
-	 * @param { 'write' | 'read' } role
+	 * @param { Role } role
 	 * @returns {Promise<void>}
 	 */
 	async updateItemPermission(itemId, permissionIdToUpdate, role) {
@@ -184,23 +185,25 @@ export class SharePointDrive {
 	 *
 	 * @param { string } itemId
 	 * @param { { id: string, email: string } } user
-	 * @param { 'write' | 'read' } permission
+	 * @param { Role } permission
 	 * @returns { Promise<void> }
 	 */
 	async upsertItemUsersPermission(itemId, user, permission) {
 		const itemsPermissions = await this.getItemPermissions(itemId);
 		if (itemsPermissions.length > 0) {
-			const findId = (id) =>
-				itemsPermissions.findIndex((permission) => permission.grantedToV2?.user.id.toString() === id);
+			const itemPermission = itemsPermissions.find(
+				(permission) => permission.grantedToV2?.user.id.toString() === user.id.toString()
+			);
 
-			const idIndex = findId(user.id);
-			if (idIndex >= 0) {
-				if (itemsPermissions[idIndex].roles[0] !== permission) {
-					await this.updateItemPermission(itemId, itemsPermissions[idIndex].id.toString(), permission);
+			if (itemPermission) {
+				if (itemPermission.roles[0] !== permission) {
+					await this.updateItemPermission(itemId, itemPermission.id.toString(), permission);
 				}
 			} else {
 				await this.addItemPermissions(itemId, { users: [user], role: permission });
 			}
+		} else {
+			await this.addItemPermissions(itemId, { users: [user], role: permission });
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


packages/sharepoint/src/sharepoint/drives/drives.js
```diff
@@ -93,4 +94,142 @@ export class SharePointDrive {
 			throw e.statusCode ?? e;
 		}
 	}
+
+	/**
+	 *
+	 * @param itemId
+	 * @returns {Promise<*>}
```
AlexMoulin, line 101: Set type of return to list of permissions (use fixture to get accurate sample)

packages/sharepoint/src/sharepoint/drives/drives.js
```diff
@@ -93,4 +94,142 @@ export class SharePointDrive {
...
+
+		await this.client.api(urlBuilder.toString()).post(permission);
+	}
+
+	/**
+	 *
+	 * @param { string } itemId
+	 * @param { Array<{ id: string, email: string }> } users
+	 * @param { 'write' | 'read' } permission
+	 * @returns { Promise<void> }
```
AlexMoulin, line 179: I did have a system where it would update each permission 1 by 1 so it would list the successes and failures (not sure if that is better?) **Feedback appreciated**

AlexMoulin: The structure I tried was: 
`
let response = {
added: { success: [email: ... ], failure: [{email: ..., reason: ... }]},
updated: { success: [email: ... ], failure: [{email: ..., reason: ... }]},
ignored: [email: ...]
`

beejjacobs: I think some change in the API is needed, since the changes are applied in a loop and not atomic, so some permission changes may have succeeded even if a later one errors.

Perhaps if update only accepts one user, then this upsert method should just do one user too? Or you return a response per use as you've suggested.

packages/sharepoint/src/sharepoint/drives/drives.js
```diff
@@ -87,10 +89,136 @@ export class SharePointDrive {
...
-			await this.client.api(urlBuilder.toString()).post(newDriveItem);
-		} catch (e) {
-			throw e.statusCode ?? e;
+		await this.client.api(urlBuilder.toString()).post(newDriveItem);
+	}
+
+	/**
+	 *
+	 * @param itemId
+	 * @returns {Promise<ListItemPermissionsResponse>}
```
beejjacobs, line 98: I don't think this return type is quite right, as you're returning the value property rather than the whole response?

packages/sharepoint/src/sharepoint/drives/drives.js
```diff
@@ -87,10 +89,136 @@ export class SharePointDrive {
...
+	}
+
+	/**
+	 *
+	 * @param {string} itemId Id of the folder to share
+	 * @param {boolean} [requireSignIn] Specifies whether the recipient of the invitation is required to sign in to view the shared item. Defaults to true
+	 * @param {boolean} [sendInvitation] 	If true, a sharing link is sent to the recipient. Otherwise, a permission is granted directly without sending a notification. Defaults to false
+	 * @param { 'write' | 'read' } role
+	 * @param { Array<{ id: string, email: string }>} users
+	 * @param {string} [message] Message that accompanies the Invitation if sent (defaults to '')
```
beejjacobs, line 120: I think WebStorm doesn't complain about this for some reason, but these are all properties of the second parameter so the docs should reflect that, e.g.

```
@param {Object} options
@param {boolean} [options.requireSignIn]
...
```

packages/sharepoint/src/sharepoint/drives/drives.js
```diff
@@ -87,10 +89,136 @@ export class SharePointDrive {
...
+
+		return response.value;
+	}
+
+	/**
+	 *
+	 * @param {string} itemId Id of the folder to share
+	 * @param {boolean} [requireSignIn] Specifies whether the recipient of the invitation is required to sign in to view the shared item. Defaults to true
+	 * @param {boolean} [sendInvitation] 	If true, a sharing link is sent to the recipient. Otherwise, a permission is granted directly without sending a notification. Defaults to false
+	 * @param { 'write' | 'read' } role
```
beejjacobs, line 118: The code seems to also allow 'owner'?

packages/sharepoint/src/sharepoint/drives/drives.js
```diff
@@ -87,10 +89,136 @@ export class SharePointDrive {
...
+		};
+
+		await this.client.api(urlBuilder.toString()).post(permission);
+	}
+
+	/**
+	 *
+	 * @param {string} itemId
+	 * @param {string} permissionIdToUpdate
+	 * @param { 'write' | 'read' } role
```
beejjacobs, line 160: Perhaps we should have a shared definition for the 'role' type?

packages/sharepoint/src/sharepoint/drives/drives.js
```diff
@@ -87,10 +89,136 @@ export class SharePointDrive {
...
+	/**
+	 *
+	 * @param { string } itemId
+	 * @param { { id: string, email: string } } user
+	 * @param { 'write' | 'read' } permission
+	 * @returns { Promise<void> }
+	 */
+	async upsertItemUsersPermission(itemId, user, permission) {
+		const itemsPermissions = await this.getItemPermissions(itemId);
+		if (itemsPermissions.length > 0) {
```
beejjacobs, line 192: With this if, we wouldn't add item permissions if there were no existing ones? is that expected?

packages/sharepoint/src/sharepoint/drives/drives.js
```diff
@@ -87,10 +89,136 @@ export class SharePointDrive {
...
+			const findId = (id) =>
+				itemsPermissions.findIndex((permission) => permission.grantedToV2?.user.id.toString() === id);
+
+			const idIndex = findId(user.id);
+			if (idIndex >= 0) {
+				if (itemsPermissions[idIndex].roles[0] !== permission) {
```
beejjacobs, line 198: Nothing wrong here, but this `findId` function is used once, so may be overkill, and we don't need the index particularly. Could be simplified to:

```
const itemPermission = itemPermissions.find((permission) => permission.grantedToV2?.user.id.toString() === id);
if (itemPermission) {
  if (itemPermission.roles[0]  !== permission)
...
```

What do you think?

packages/sharepoint/src/sharepoint/drives/drives.test.js
```diff
@@ -127,4 +125,394 @@ describe('drives', () => {
...
+				error.statusCode = 404;
+				throw error;
+			};
+			try {
+				await sharePointDrive.getItemPermissions(itemId);
+				assert.fail('Expected error was not thrown');
+			} catch (e) {
+				assert.equal(e.statusCode, 404);
+				assert.equal(e.message, 'The resource could not be found.');
+			}
```
beejjacobs, line 170: This looks fine - just FYI there is a 'rejects' assertion, which you may find useful:

https://nodejs.org/docs/latest-v22.x/api/assert.html#assertrejectsasyncfn-error-message
e.g.
```
await assert.rejects(
  () => updateCase({ req: mockReq, res: mockRes, data }),
  (err) => {
    assert.strictEqual(err.name, 'Error');
    assert.strictEqual(err.message, 'Error updating case (E101)');
    return true;
  }
);
```

packages/sharepoint/src/sharepoint/drives/drives.test.js
```diff
@@ -127,4 +125,394 @@ describe('drives', () => {
...
+			try {
+				await sharePointDrive.deleteItemUserPermission(itemId, '1');
+			} catch (e) {
+				assert.equal(e.statusCode, 404);
+				assert.equal(e.message, 'The resource could not be found.');
+			}
+		});
+	});
+	describe('upsertItemUsersPermission', () => {
+		test('should add the user permission if no permission is found', async () => {
```
beejjacobs, line 411: Is it worth asserting that `sharePointDrive.addItemPermissions` was called?

beejjacobs (commented):
A few comments, but mostly polish ones.... just one which is a potential bug around upserts.

packages/sharepoint/src/sharepoint/drives/drives.test.js
beejjacobs: I suppose in essence yes, but if the implementation of `addItemPermissions` were to change, this may fail - for the purposes of the upsert test you want to check that `addItemPermissions` as a method was called? This just relates to keeping tests focussed in scope and only testing the relevant method/functionality.

I don't think its a big deal here, as its all part of the same class and larger test suite!

AlexMoulin: I removed the code that was testing the functionality of add and update in the upsert method and now just check what gets called. :) Thanks!

beejjacobs (approved):

tomxelliott (approved):

PR #87, opened 2025-01-30
feat(manage): add assign write access to applicant and agent on case create

AlexMoulin:
## Describe your changes

- applicant is granted write access to their folder in the newly created folder structure
- agent (if assigned) is granted write access to the applicant folder in the newly created folder structure 

https://pins-ds.atlassian.net/browse/CROWN-96

### DIFF
--- a/apps/manage/src/app/views/cases/create-a-case/save.js
+++ b/apps/manage/src/app/views/cases/create-a-case/save.js
@@ -3,6 +3,7 @@ import { clearDataFromSession } from '@pins/dynamic-forms/src/lib/session-answer
 import { JOURNEY_ID } from './journey.js';
 import { caseReferenceToFolderName } from '@pins/crowndev-lib/util/name.js';
 import { toFloat } from '@pins/crowndev-lib/util/numbers.js';
+import { getSharePointReceivedPathForCase } from '@pins/crowndev-lib/util/sharepoint-path.js';
 
 /**
  * @param {Object} opts
@@ -47,10 +48,7 @@ export function buildSaveController({ db, logger, config, getSharePointDrive }) 
 				'SharePoint not enabled, to use SharePoint functionality setup SharePoint environment variables. See README'
 			);
 		} else {
-			await sharePointDrive.copyDriveItem({
-				copyItemId: config.sharePoint.caseTemplateId,
-				newItemName: caseReferenceToFolderName(reference)
-			});
+			await createCaseSharePointActions(sharePointDrive, config, caseReferenceToFolderName(reference), answers);
 		}
 		// todo: redirect to check-your-answers on failure?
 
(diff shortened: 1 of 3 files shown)


packages/lib/util/sharepointPath.js
```diff
@@ -0,0 +1,15 @@
...
+ * @param {'Received' | 'Published' | 'Documentation'} [type]
+ * @param {'Applicant' | 'LPA' } [user] Requires type = Received
+ * @returns {string}
+ */
+export function getSharePointPathForCase({ caseReference, type, user }) {
+	if (type !== 'Received' && user) {
+		throw Error('Invalid path');
+	}
+	return `${caseReferenceToFolderName(caseReference)}${type ? '/' + type : ''}${user && type === 'Received' ? '/' + user : ''}`;
+}
```
beejjacobs, line 15: I think this function is perhaps too generic, its a little confusing to read for me.

What if we had one specifically for 'received' folders only?

apps/manage/src/app/views/cases/create-a-case/save.js
```diff
@@ -51,6 +52,31 @@ export function buildSaveController({ db, logger, config, getSharePointDrive })
...
+				//lpaReceivedFolder
+			] = await Promise.all([
+				sharePointDrive.getItemsByPath(applicantReceivedPath, [])
+				//sharePointDrive.getItemsByPath(lpaReceivedPath, [])
+			]);
+
+			let users = [{ email: answers.applicantEmail, id: '' }];
+			if (hasAnswers(answers, 'agent')) {
+				users.push({ email: answers.agentEmail, id: '' });
+			}
```
beejjacobs, line 77: Applicant email is optional if there's an agent, so I think it'd be more robust if we just explicitly check if `agentEmail` and `applicantEmail` have values, and add them if they do.

apps/manage/src/app/views/cases/create-a-case/save.js
```diff
@@ -51,6 +52,31 @@ export function buildSaveController({ db, logger, config, getSharePointDrive })
...
+			// 	});
+			const [
+				applicantReceivedFolder
+				//lpaReceivedFolder
+			] = await Promise.all([
+				sharePointDrive.getItemsByPath(applicantReceivedPath, [])
+				//sharePointDrive.getItemsByPath(lpaReceivedPath, [])
+			]);
+
+			let users = [{ email: answers.applicantEmail, id: '' }];
```
beejjacobs, line 74: This can be const

beejjacobs (commented):
Looks pretty good, but I think is starting to add quite a lot to 'save' - perhaps a small service method is required here to encapsulate the SharePoint actions needed?

e.g it could take the sharePointDrive instance, case reference, and answers, and do all the SharePoint steps.

What do you reckon?

apps/manage/src/app/views/cases/create-a-case/save.js
```diff
@@ -216,3 +214,54 @@ function idFromReference(reference = '') {
...
+		applicantReceivedFolder
+		//lpaReceivedFolder
+	] = await Promise.all([
+		sharePointDrive.getItemsByPath(applicantReceivedPath, [])
+		//sharePointDrive.getItemsByPath(lpaReceivedPath, [])
+	]);
+
+	const users = [];
+	hasAnswers(answers, 'applicant') ? users.push({ email: answers.applicantEmail, id: '' }) : null;
+	hasAnswers(answers, 'agent') ? users.push({ email: answers.agentEmail, id: '' }) : null;
```
beejjacobs, line 245: This is partly preference, but I prefer to avoid doing things in a ternary, I think "if" is clearer for that, especially when in the false case we don't want to do anything.

Secondly, we should check `answers.applicantEmail` explicitly - if any applicant* answer is filled in, hasAnswers will return true, but that doesn't mean email is populated.

apps/manage/src/app/views/cases/create-a-case/save.js
```diff
@@ -47,10 +48,7 @@ export function buildSaveController({ db, logger, config, getSharePointDrive })
 				'SharePoint not enabled, to use SharePoint functionality setup SharePoint environment variables. See README'
 			);
 		} else {
-			await sharePointDrive.copyDriveItem({
-				copyItemId: config.sharePoint.caseTemplateId,
-				newItemName: caseReferenceToFolderName(reference)
-			});
+			await createCaseSharePointActions(sharePointDrive, config, caseReferenceToFolderName(reference), answers);
```
beejjacobs, line 51: Much neater - love it!

tomxelliott (approved):

PR #91, opened 2025-01-31
Feat/case involvement

beejjacobs:
## Describe your changes

Read Entra Groups for case involvement fields

https://pins-ds.atlassian.net/browse/CROWN-75
https://pins-ds.atlassian.net/browse/CROWN-86

Shows display name for the value:

image

and uses the select component for editing:

image


apps/manage/src/app/app.js
```diff
@@ -34,7 +39,7 @@ export function getApp(config, logger) {
 
 	const sessionMiddleware = getSessionMiddleware({
 		redis,
-		secure: config.NODE_ENV === 'production',
+		secure: false, // config.NODE_ENV === 'production',
```
tomxelliott, line 42: do we want to commit this change?

tomxelliott (dismissed):

apps/manage/src/app/app.js
beejjacobs: Good spot, will revert that bit!

tomxelliott (approved):

PR #104, opened 2025-02-06
fix(manage): fix bug that causes permission process to not find folderId

AlexMoulin:
## Describe your changes
Fix call to path was not returning folderId

### DIFF
--- a/packages/lib/util/sharepoint-path.js
+++ b/packages/lib/util/sharepoint-path.js
@@ -1,12 +1,22 @@
 /**
  *
+ * @param {SharePointDrive} sharePointDrive
  * @param {string} caseRootName
  * @param {'Applicant' | 'LPA' } [user] Requires type = Received
- * @returns {string}
+ * @returns {Promise<string>}
  */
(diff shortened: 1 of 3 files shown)


packages/lib/util/sharepoint-path.js
```diff
@@ -1,12 +1,22 @@
...
  */
-export function getSharePointReceivedPathForCase({ caseRootName, user }) {
+export async function getSharePointReceivedPathId(sharePointDrive, { caseRootName, user }) {
 	if (!user) {
-		throw Error('Invalid path');
+		throw new Error('Invalid path');
 	}
-	return `${caseRootName}'/Received/'${user}`;
+
+	const receivedFolders = await sharePointDrive.getItemsByPath(`${caseRootName}/Received`, []);
```
beejjacobs, line 13: I guess we could just select id and name, as the only two fields we need? Can be a future improvement!

beejjacobs (approved):
LGTM

tomxelliott (approved):

PR #114, opened 2025-02-10
feat(manage): update validation for date period component

tomxelliott:
## Describe your changes
Edit Case Details: Representations
update validation for date period component

https://pins-ds.atlassian.net/browse/CROWN-85


beejjacobs (approved):
Looks good. I think some of this is a bit specific for a generic date-period validator, but that is not a big deal when we've got the time pressures we do!

Also for the tests, I'd probably have left the default settings to not check for future dates, so all the tests would have remained. Then added extra tests with an explicit future tag - again not a big deal for now.

beejjacobs (approved):

PR #139, opened 2025-02-17
fix(manage): If coordinates not populated leave blank on CYA

tomxelliott:
## Describe your changes
Fix: Coordinates Check your answers says 'not started' instead of being blank
If coordinates not populated leave blank on CYA

https://pins-ds.atlassian.net/browse/CROWN-141

### DIFF
--- a/packages/dynamic-forms/src/components/multi-field-input/question.js
+++ b/packages/dynamic-forms/src/components/multi-field-input/question.js
@@ -132,7 +132,14 @@ export default class MultiFieldInputQuestion extends Question {
 			return answer ? acc + (field.formatPrefix || '') + answer + (field.formatJoinString || '\n') : acc;
 		}, '');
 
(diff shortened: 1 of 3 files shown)


packages/dynamic-forms/src/components/multi-field-input/question.js
beejjacobs, line 1: For this purpose in other places, I added text overrides. I think we just need to add support for this in multi-field-input?

e.g. see https://github.com/Planning-Inspectorate/crown-developments/blob/adaefa430540dd51d8382f4a4f34b2239410c51d/apps/manage/src/app/views/cases/view/questions.js#L394

beejjacobs (approved):

beejjacobs (commented):
LGTM!

beejjacobs (approved):
LGTM

PR #150, opened 2025-02-20
feat(dynamic-forms): add text-entry-redact component

beejjacobs:
## Describe your changes

Dynamic forms component for text redaction - to support manage reps

N/A

### DIFF
--- a/packages/dynamic-forms/src/components/text-entry-redact/index.njk
+++ b/packages/dynamic-forms/src/components/text-entry-redact/index.njk
@@ -1,0 +1,166 @@
+{% extends layoutTemplate %}
+
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
+{% from "govuk/components/textarea/macro.njk" import govukTextarea %}
+{% from "govuk/components/label/macro.njk" import govukLabel %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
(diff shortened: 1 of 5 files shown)


packages/dynamic-forms/src/components/text-entry-redact/index.njk
```diff
@@ -0,0 +1,166 @@
...
+                        classes: "govuk-button--secondary",
+                        id: "redact-undo-all",
+                        type: "button"
+                    }) }}
+                </div>
+
+                {{ govukButton({
+                    text: continueButtonText,
+                    type: "submit",
+                    attributes: { "data-cy":"button-save-and-continue"}
```
tomxelliott, line 97: super minor comment - do we want to add spacing to keep consistent with other spacing on json objects, like below?
`                    attributes: { "data-cy": "button-save-and-continue" }
`

tomxelliott (approved):
looks good, just one really small comment

packages/dynamic-forms/src/components/text-entry-redact/index.njk
beejjacobs: yes sounds good - annoyingly I don't think prettier supports nunjucks so the formatting is a bit hit-and-miss!

tomxelliott (approved):

PR #156, opened 2025-02-21
feat(package): add publish case functionality

AlexMoulin:
## Describe your changes
**Package updates:**
- Add date utils to check if date is past, future or today with tests
- Update SessionClear to allow for arrays of fields

**Case view list updates:**
- Update cases/view view.njk to add publish and unpublish buttons based on publishDate with tests
- Add validation when clicking publish button
- Update cases/view questions to make publishDate uneditable
- Update question utils to allow for dateQuestion to process editable rows

**Publish case updates:**

- Add success page for publishing
- Add publish route
- update DB with publishDate
- display success page

https://pins-ds.atlassian.net/browse/CROWN-109
https://pins-ds.atlassian.net/browse/CROWN-91

### DIFF
--- a/apps/manage/src/app/views/cases/view/controller.test.js
+++ b/apps/manage/src/app/views/cases/view/controller.test.js
@@ -9,6 +9,7 @@ import { Prisma } from '@prisma/client';
 describe('case details', () => {
 	const mockGetEntraClient = mock.fn(() => null);
 	const groupIds = { caseOfficer: 'id-1', inspector: 'id-2' };
+	mock.timers.enable({ apis: ['Date'], now: new Date('2025-01-01T03:24:00') });
 
 	describe('buildGetJourneyMiddleware', () => {
 		it('should throw if no id', async () => {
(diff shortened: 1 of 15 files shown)


packages/dynamic-forms/src/lib/date-utils.js
```diff
@@ -77,3 +77,27 @@ export function endOfDay() {
...
+ */
+export const dateIsInTheFuture = (date) => {
+	return isAfter(date, endOfDay());
+};
+
+/**
+ * @param {Date} date
+ * @returns {boolean}
+ */
+export const dateIsInThePast = (date) => {
```
beejjacobs, line 93: This isn't strictly "in the past" but more like "before today" isn't it?  might worth clarifying the naming of these

apps/manage/src/app/views/cases/view/publish/controller.js
```diff
@@ -0,0 +1,131 @@
...
+				AgentContact: { include: { Address: true } },
+				Lpa: { include: { Address: true } }
+			}
+		});
+
+		if (!crownDevelopment) {
+			return notFoundHandler(req, res);
+		}
+
+		const answers = {
```
beejjacobs, line 68: This could just be an array - not sure that the object keys here are necessary - but no harm with the current format.

apps/manage/src/app/views/cases/view/publish/controller.js
```diff
@@ -0,0 +1,131 @@
...
+			}
+		}
+
+		if (errors.length > 0) {
+			addSessionData(req, id, { errors });
+			return res.redirect(`/cases/${id}`);
+		}
+
+		const reference = crownDevelopment.reference;
+		addSessionData(req, id, { reference });
```
beejjacobs, line 103: It'd be good to keep any session data specific, perhaps publishErrors would be a better field name here?

and for success, perhaps `casePublished: true` or similar? Just saving the reference in session doesn't have much meaning.

apps/manage/src/app/views/cases/view/publish/controller.test.js
```diff
@@ -0,0 +1,227 @@
...
+			const mockDb = {
+				crownDevelopment: {
+					findUnique: mock.fn(() => ({ id: 'case-1' }))
+				}
+			};
+			const middleware = buildGetValidatedCaseMiddleware({ db: mockDb, logger: mockLogger() });
+
+			await middleware(mockReq, mockRes, next);
+			assert.strictEqual(mockDb.crownDevelopment.findUnique.mock.calls.length, 1);
+			console.log(mockDb.crownDevelopment.findUnique.mock.calls[0][0]);
```
beejjacobs, line 42: Is this meant to be here?

apps/manage/src/app/views/cases/view/publish/controller.test.js
```diff
@@ -0,0 +1,227 @@
...
+			const next = mock.fn();
+			const mockDb = {
+				crownDevelopment: {
+					findUnique: mock.fn(() => ({ id: 'case-1' }))
+				}
+			};
+			const middleware = buildGetValidatedCaseMiddleware({ db: mockDb, logger: mockLogger() });
+
+			await middleware(mockReq, mockRes, next);
+			assert.strictEqual(mockDb.crownDevelopment.findUnique.mock.calls.length, 1);
```
beejjacobs, line 41: you should use `mock.callCount()` instead of checking calls length - its slightly more efficient and the recommended way (see https://nodejs.org/docs/latest-v22.x/api/test.html#class-mockfunctioncontext)

apps/manage/src/app/views/cases/view/controller.test.js
```diff
@@ -170,6 +170,144 @@ describe('case details', () => {
...
+				render: mock.fn((view, data) => nunjucks.render(view, data))
+			};
+			const mockReq = {
+				params: { id: 'case-1' },
+				baseUrl: 'case-1',
+				session: {}
+			};
+			const mockDb = {
+				crownDevelopment: {
+					findUnique: mock.fn(() => ({ id: 'case-1', name: 'Case 1', publishDate: new Date('1995-12-17T03:24:00') }))
```
beejjacobs, line 188: I've not actually used it yet - but turns out you can mock Date quite easily, so we set in the test when 'now' is for dates like this. This may also be useful for your 'publishDate is in the future' test

https://nodejs.org/docs/latest-v22.x/api/test.html#dates

packages/dynamic-forms/src/lib/date-utils.test.js
```diff
@@ -54,4 +54,62 @@ describe('format-date', () => {
 			});
 		}
 	});
+	describe('dateIsInTheFuture', () => {
+		const tests = [
+			{
+				input: new Date('2090-02-20T15:00Z'),
```
beejjacobs, line 60: see other note on Date mocking - may be useful to try!

beejjacobs (changes requested):
Functionality all looks good, just a few comments to consider for tidiness/consistency

apps/manage/src/app/views/cases/view/publish/controller.test.js
AlexMoulin: No, nice catch.

apps/manage/src/app/views/cases/view/publish/controller.js
AlexMoulin: That makes sense. For success though, the reason why I passed reference was to avoid calling the database again just to get the reference for the success banner?

beejjacobs: That seems sensible - we can save both then?

`addSessionData(req, id, { reference, casePublished: true });`

tomxelliott (dismissed):

apps/manage/src/app/views/cases/view/publish/controller.js
```diff
@@ -0,0 +1,131 @@
...
+	const reference = readSessionData(req, id, 'reference');
+	if (!reference) {
+		throw new Error('invalid publish case session');
+	}
+
+	const data = {
+		id,
+		reference
+	};
+	clearSessionData(req, id, 'reference');
```
beejjacobs, line 126: This should read (and clear) `casePublished` property too I reckon?

tomxelliott (approved):

packages/lib/util/session.js
```diff
@@ -69,13 +69,21 @@ export function readSessionData(req, id, field, defaultValue, sessionField = 'ca
...
  * @param {string} [sessionField]
  */
-export function clearSessionData(req, id, field, sessionField = 'cases') {
+export function clearSessionData(req, id, fieldOrFields, sessionField = 'cases') {
 	if (!req.session) {
 		return; // no need to error here
 	}
+	if (fieldOrFields instanceof Array) {
+		fieldOrFields.forEach((field) => {
+			const fieldProps = (req.session[sessionField] && req.session[sessionField][id]) || {};
```
beejjacobs, line 81: Super minor, but this is repeated now and could be pulled up to e.g. after line 78

beejjacobs (approved):

packages/lib/util/session.js
AlexMoulin: I'll update it in the unpublish code, it uses the session data too.

PR #178, opened 2025-02-28
feat/manage-representations

beejjacobs:
## Describe your changes

Written representations dashboard with comments grouped by status into tabs

https://pins-ds.atlassian.net/browse/CROWN-144

### DIFF
--- a/apps/manage/src/app/views/cases/view/manage-reps/index.js
+++ b/apps/manage/src/app/views/cases/view/manage-reps/index.js
@@ -1,0 +1,23 @@
+import { Router as createRouter } from 'express';
+import { asyncHandler } from '@pins/crowndev-lib/util/async-handler.js';
+import { buildListReps } from './list/controller.js';
+import { addRep } from './add/controller.js';
+import { viewRep } from './view/controller.js';
+import { reviewRep } from './review/controller.js';
+
+/**
+ * @param {Object} opts
+ * @param {import('@prisma/client').PrismaClient} opts.db
+ * @returns {import('express').Router}
+ */
+export function createRoutes({ db }) {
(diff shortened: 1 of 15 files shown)


apps/manage/src/app/views/cases/view/manage-reps/view/controller.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Placeholder
+ * @type {import('express').Handler}
+ */
+export function viewRep(req, res) {
+	res.render('views/cases/view/manage-reps/view/view.njk', {
+		backLink: '#',
```
tomxelliott, line 7: is this still to be determined?

apps/manage/src/app/views/cases/view/manage-reps/index.js
```diff
@@ -0,0 +1,23 @@
...
+ * @param {import('@prisma/client').PrismaClient} opts.db
+ * @returns {import('express').Router}
+ */
+export function createRoutes({ db }) {
+	const router = createRouter({ mergeParams: true });
+	const list = buildListReps({ db });
+
+	router.get('/', asyncHandler(list));
+	router.get('/add', addRep);
+	router.get('/:repId/view', viewRep);
```
tomxelliott, line 19: is it worth making this representationId so it is clearer that it is representation rather than representative?

tomxelliott (approved):

apps/manage/src/app/views/cases/view/manage-reps/index.js
beejjacobs: Yes good thought - I will address as part of CROWN-164

PR #197, opened 2025-03-06
fix(portal): display error for max length ahead of regex error

tomxelliott:
## Describe your changes
Display error for max length ahead of regex error

https://pins-ds.atlassian.net/browse/CROWN-310


beejjacobs (approved):
Future improvement would be to allow configuring the order so its generic, but this is fine for now.

PR #215, opened 2025-03-10
feat(portal): begin configuration of written reps page

tomxelliott:
## Describe your changes
[Portal] View Written Representations Page

https://pins-ds.atlassian.net/browse/CROWN-229

Screenshot 2025-03-11 at 16 27 04

Screenshot 2025-03-11 at 16 27 45

Screenshot 2025-03-11 at 16 27 58

Screenshot 2025-03-11 at 16 35 24

### DIFF
--- a/apps/portal/src/app/views/applications/view/written-representations/controller.js
+++ b/apps/portal/src/app/views/applications/view/written-representations/controller.js
@@ -66,8 +66,7 @@ export function buildWrittenRepresentationsPage({ db }) {
 			representations: representations.map(representationToViewModel),
 			numberOfRepresentations: representations.length,
 			links: applicationLinks(id, haveYourSayPeriod, representationsPublishDate),
-			currentUrl: req.originalUrl,
-			selectedResultsToShow: req.query.resultsToShow || 25
+			currentUrl: req.originalUrl
 		});
 	};
 }
--- a/apps/portal/src/app/views/applications/view/written-representations/controller.js
+++ b/apps/portal/src/app/views/applications/view/written-representations/controller.js
@@ -29,30 +29,46 @@ export function buildWrittenRepresentationsPage({ db }) {
 			return notFoundHandler(req, res);
 		}
 
-		const representations = await db.representation.findMany({
-			where: {
-				// applicationId: id,
-				statusId: REPRESENTATION_STATUS_ID.ACCEPTED
-			},
-			select: {
-				reference: true,
-				submittedDate: true,
-				comment: true,
-				commentRedacted: true,
-				submittedByAgentOrgName: true,
-				submittedForId: true,
-				representedTypeId: true,
-				containsAttachments: true,
-				SubmittedFor: { select: { displayName: true } },
-				SubmittedByContact: { select: { fullName: true, isAdult: true } },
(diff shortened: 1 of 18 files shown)


apps/portal/src/app/views/applications/view/written-representations/controller.js
```diff
@@ -28,6 +29,47 @@ export function buildWrittenRepresentationsPage({ db }) {
 			return notFoundHandler(req, res);
 		}
 
+		const selectedItemsPerPage = Number(req.query?.itemsPerPage) || 25;
+		const pageNumber = Math.max(1, Number(req.query?.page) || 1);
+		const pageSize = [25, 50, 100].includes(selectedItemsPerPage) ? selectedItemsPerPage : 100;
+		const skipSize = (pageNumber - 1) * pageSize;
+
+		const [representations, totalRepresentations] = await Promise.all([
```
AlexMoulin, line 37: Might be worth having some error handling around the db query? If the db is unreachable, would it give a NaN to your Math?

AlexMoulin (changes requested):
Look good overall, one comment from me :)

apps/portal/src/app/views/applications/view/written-representations/controller.js
beejjacobs: We need to wrap this query so we don't expose Prisma errors to the users if something does fail - see `wrapPrismaError` as a starting point.

beejjacobs (commented):
Looks good, I'd just like to see the test coverage improved for the pagination logic; and Alex's comment about error handling for DB calls does need looking at please.

apps/portal/src/app/views/applications/view/written-representations/view.njk
```diff
@@ -21,7 +25,19 @@
...
 			</div>
+
+			<div id="pagination" class="govuk-body">
+				{{ paginationOptions.renderPagination(selectedItemsPerPage, totalRepresentations, resultsStartNumber, resultsEndNumber) }}
+
+				<hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+
+				{% for i in range(representations.length if representations.length < selectedItemsPerPage else selectedItemsPerPage) %}
+					{{ representation.renderRepresentation(representations[i]) }}
+				{% endfor %}
```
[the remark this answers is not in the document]
tomxelliott: yeah, you're right. this is a holdover from my initial implementation using client side js. will use `{% for rep in representations %}` instead

apps/portal/src/app/views/applications/view/written-representations/pagination.test.js
beejjacobs, line 1: Love this - we could do with more tests like this across PINs for all the nunjuck macros we have floating about!

apps/portal/src/app/views/applications/view/written-representations/controller.test.js
beejjacobs, line 1: Looks good - and I like how because pagination is tested separate we can just check the view data values for this, to check mapping from query params.

beejjacobs (approved):
Comments addressed, test coverage looks good. As per Teams `read-mode` section not considered as that will be a separate PR.

AlexMoulin (approved):

PR #223, opened 2025-03-14
feat(manage): add add-rep journey for myself

AlexMoulin:
## Describe your changes
Add the myself add representation journey
https://pins-ds.atlassian.net/browse/CROWN-269

image
image
image
image
image
image
image
image


packages/lib/forms/representations/sections.js
```diff
@@ -56,6 +63,29 @@ function myselfSection(questions) {
 		.addQuestion(questions.myselfTellUsAboutApplication);
 }
 
+/**
+ * Myself section for the add representation journey (contains more questions that the default section)
+ *
+ * @param {Questions} questions
+ * @returns {Section}
+ */
+function addRepMyselfSection(questions) {
```
tomxelliott, line 72: is it worth adding these add rep sections into their own file to more clearly distinguish between the have your say and add rep stuff?

packages/dynamic-forms/src/validator/address-validator.test.js
```diff
@@ -50,6 +50,8 @@ describe('AddressValidator', () => {
 		assert.deepStrictEqual(errors, {});
 	});
 
+	it('should validate address with errors when required fields are missing', async () => {});
```
tomxelliott, line 53: are we expecting a test here?

packages/lib/forms/representations/types.d.ts
```diff
@@ -39,6 +42,8 @@ export namespace HaveYourSay {
 	export interface Myself {
 		myselfIsAdult: boolean;
 		myselfFullName?: string;
+		myselfAddress?: string;
```
beejjacobs, line 45: This'd be of type `Address` from `import { Address } from '@pins/dynamic-forms/src/lib/address.js';` I think, not just a string.

packages/dynamic-forms/src/validator/address-validator.js
```diff
@@ -33,9 +33,12 @@ export default class AddressValidator extends BaseValidator {
 	 * creates an instance of an AddressValidator
 	 * @param {Object} opts
 	 * @param {boolean} [opts.required]
+	 * @param {{[key: string], value: boolean}} [opts.requiredFields]
```
beejjacobs, line 36: This isn't quite right, this is an object like:
```
{
  value: true,
  anyKey: 'anyValue'
}
```

You perhaps meant `Object<string, boolean>` - so all keys are strings, and all values are booleans. Or in the same syntax you have here, that would be:

`{[key: string]: boolean}`

beejjacobs (commented):
Other comments addressed, looks good. Two more minor ones!

beejjacobs (approved):
All addressed, looks good.

PR #226, opened 2025-03-14
chore(seed): allow additional written reps to be generated

tomxelliott:


package.json
```diff
@@ -38,6 +38,7 @@
     "express": "^4.21.2",
     "express-session": "^1.18.1",
     "express-validator": "^7.2.1",
+    "@faker-js/faker": "^9.6.0",
```
beejjacobs, line 41: This can be a dev-dep I think

beejjacobs (commented):
Looks good, just a small changes to deps

beejjacobs (approved):

PR #241, opened 2025-03-18
feat(manage): AUTOMATIC NOTIFICATION: To applicant - acknowledge pre-notification

tomxelliott:
## Describe your changes
AUTOMATIC NOTIFICATION: To applicant - acknowledge pre-notification

https://pins-ds.atlassian.net/browse/CROWN-244
https://pins-ds.atlassian.net/browse/CROWN-245

### DIFF
--- a/apps/manage/src/app/config-types.d.ts
+++ b/apps/manage/src/app/config-types.d.ts
@@ -43,5 +43,10 @@ interface Config {
 		disabled: boolean;
 		apiKey: string;
 		testTemplate: string;
+		preAckTemplate: string;
+		templates: {
+			test: string;
+			acknowledgePreNotification: string;
+		};
 	};
 }
--- a/apps/manage/src/app/views/cases/create-a-case/save.js
+++ b/apps/manage/src/app/views/cases/create-a-case/save.js
@@ -3,6 +3,7 @@ import { clearDataFromSession } from '@pins/dynamic-forms/src/lib/session-answer
 import { JOURNEY_ID } from './journey.js';
 import { toFloat } from '@pins/crowndev-lib/util/numbers.js';
 import { caseReferenceToFolderName, getSharePointReceivedPathId } from '@pins/crowndev-lib/util/sharepoint-path.js';
+import { yesNoToBoolean } from '@pins/dynamic-forms/src/components/boolean/question.js';
 
 /**
  * @param {Object} opts
@@ -10,9 +11,10 @@ import { caseReferenceToFolderName, getSharePointReceivedPathId } from '@pins/cr
  * @param {import('pino').BaseLogger} logger
  * @param {import('../../../config-types.js').Config} config
  * @param {function(session): SharePointDrive} getSharePointDrive
+ * @param {import('@pins/crowndev-lib/govnotify/gov-notify-client').GovNotifyClient|null} govNotifyClient
  * @returns {import('express').Handler}
  */
-export function buildSaveController({ db, logger, config, getSharePointDrive }) {
+export function buildSaveController({ db, logger, config, getSharePointDrive, govNotifyClient }) {
 	return async (req, res) => {
 		if (!res.locals || !res.locals.journeyResponse) {
 			throw new Error('journey response required');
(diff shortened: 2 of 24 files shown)


apps/manage/src/app/config-types.d.ts
```diff
@@ -43,5 +43,6 @@ interface Config {
 		disabled: boolean;
 		apiKey: string;
 		testTemplate: string;
+		preAckTemplate: string;
```
AlexMoulin, line 46: Not sure about the naming convention, there will be quite a few templates. But I can't think  of a good short name for it maybe acknowledgePreNotificationTemplate, more verbose but much clearer. Alternatively, we could use comments but that wouldn't carry over each time it is used.

AlexMoulin (dismissed):
Really minor comment, more about clarity than anything.

AlexMoulin (approved):

apps/manage/src/app/config-types.d.ts
beejjacobs: I wonder if we add a `templateIds` object, then we can have shorter names in that object.

```
govNotify: {
  disabled: boolean;
  apiKey: string;
  templateIds: {
    acknowledgePreNotification: string;
    // etc...
  }
```

packages/lib/govnotify/index.js
```diff
@@ -7,7 +7,7 @@ import { GovNotifyClient } from './gov-notify-client.js';
 let govNotify;
 
 /**
- * @param {string} [govNotifyApiKey] - Gov Notify API key
+ * @param {import('../app/config-types.js').Config} config
```
beejjacobs, line 10: I don't think that types path is right - better this just has a basic object definition here rather than pulling in a specific config type. e.g. `{disabed: boolean, apiKey: string}` - then when we call it we pass in `config.govNotify`

apps/manage/src/app/views/cases/create-a-case/save.js
```diff
@@ -41,16 +42,38 @@ export function buildSaveController({ db, logger, config, getSharePointDrive })
...
 		if (sharePointDrive === null) {
 			logger.warn(
 				'SharePoint not enabled, to use SharePoint functionality setup SharePoint environment variables. See README'
 			);
 		} else {
 			await createCaseSharePointActions(sharePointDrive, config, caseReferenceToFolderName(reference), answers);
+			const item = await sharePointDrive.getDriveItemByPath(caseReferenceToFolderName(reference), [
+				['$select', 'webUrl']
+			]);
+			sharePointLink = item.webUrl;
```
beejjacobs, line 56: I don't think this is the right link - Alex may be able to advise. We don't want the webUrl but the specific invite URL for the specific user/email address.

apps/manage/src/app/views/cases/create-a-case/save.js
```diff
@@ -41,16 +42,38 @@ export function buildSaveController({ db, logger, config, getSharePointDrive })
...
+			logger.warn(
+				'Gov Notify is not enabled, to use Gov Notify functionality setup Gov Notify environment variables. See README'
+			);
+		} else {
+			try {
+				const { hasAgent, agentEmail, applicantEmail } = req.session?.forms[JOURNEY_ID] || {};
+				const recipientEmail = hasAgent ? agentEmail : applicantEmail;
+				const params = { reference, 'sharepoint link': sharePointLink };
+
+				await govNotifyClient.sendEmail(config.govNotify.preAckTemplate, recipientEmail, { personalisation: params });
```
beejjacobs, line 70: `createCaseSharePointActions` could return the email address and the link, since it already has this logic for which email to use - I think that'd be cleaner and remove this duplicate logic.

Also, should we rename all the personalisation fields so they are camelCase? Having 'sharepoint link' with a space should work, but doesn't fit our other conventions.

apps/manage/src/app/views/cases/create-a-case/save.js
```diff
@@ -41,16 +42,38 @@ export function buildSaveController({ db, logger, config, getSharePointDrive })
...
+			} catch (error) {
+				logger.error('error dispatching email notification', error);
```
beejjacobs, line 72: For pino you need to include any object first, so `logger.error({error}, 'message'})` - and it'd be good to include some context in the message else when looking at the log you wouldn't know what it was for.  Perhaps `reference` is sufficient so we know which case it is.

packages/dynamic-forms/src/components/task-list/index.njk
```diff
@@ -36,9 +36,9 @@
     <div class="govuk-grid-row">
         <div class="govuk-grid-column-full">
             {% if summaryListData.sections | length > 1 %}
-                {{ multiTaskList(summaryListData.sections, journeyComplete) }}
+                {{ multiTaskList(summaryListData.sections, journeyComplete, showNotifyWarning) }}
             {% elif summaryListData.sections | length == 1 %}
-                {{ singleTaskList(summaryListData.sections | first, journeyComplete) }}
+                {{ singleTaskList(summaryListData.sections | first, journeyComplete, showNotifyWarning) }}
```
beejjacobs, line 41: It'd be good to handle this without modifying the dynamic-forms with this custom code and hard-coded messages (at some point this will be a separate package).

I think the solution to try is a block? e.g. `beforeSubmitButton`

```
{% if journeyComplete %}
                        {% block beforeSubmitButton %}
                        {% endblock %}
                        <form action method="post">
                            <input type="hidden" name="_csrf" value="{{_csrf}}">
                            {{ govukButton({
                                text: "Accept & Submit",
                                type: "submit"
                            }) }}
                        </form>
                        ```

The layout used for create-a-case is the `forms-check-your-answers.njk` layout, so we can then override that block there.

beejjacobs (changes requested):
There's a few bits here that need sorting please. Most pressing is that I don't think this is the right URL that we should be sending. The rest are more about tidyness and how we make sure there are good patterns going forward.

beejjacobs (approved):

PR #290, opened 2025-03-31
feat(portal): configure email sent at end of have your say journey

tomxelliott:
## Describe your changes
AUTOMATIC NOTIFICATION: Acknowledgement of representation

https://pins-ds.atlassian.net/browse/CROWN-265


apps/portal/src/app/config-types.d.ts
```diff
@@ -22,8 +22,9 @@ interface Config {
 	govNotify: {
 		disabled: boolean;
 		apiKey: string;
-		templates: {
+		templateIds: {
 			test: string;
+			acknowledgementOfRepresentation: string;
 		};
```
beejjacobs, line 28: Be good to replace this whole bit so we don't have to keep three `templateIds` definitions up to date:

```
import {NotifyConfig} from '@pins/crowndev-lib/govnotify/types';
...
govNotify: NotifyConfig;
```

Same in the manage config definition.

infrastructure/app-portal.tf
```diff
@@ -76,9 +76,10 @@ module "app_portal" {
     WEBSITE_AUTH_AAD_ALLOWED_TENANTS         = data.azurerm_client_config.current.tenant_id
 
     # gov notify
-    GOV_NOTIFY_DISABLED         = var.apps_config.gov_notify.disabled
-    GOV_NOTIFY_API_KEY          = local.key_vault_refs["crown-gov-notify-api-key"]
-    GOV_NOTIFY_TEST_TEMPLATE_ID = var.apps_config.gov_notify.templates.test_template_id
+    GOV_NOTIFY_DISABLED            = var.apps_config.gov_notify.disabled
+    GOV_NOTIFY_API_KEY             = local.key_vault_refs["crown-gov-notify-api-key"]
+    GOV_NOTIFY_TEST_TEMPLATE_ID    = var.apps_config.gov_notify.templates.test_template_id
+    GOV_NOTIFY_ACK_REP_TEMPLATE_ID = var.apps_config.gov_notify.templates.ack_rep_template_id
```
beejjacobs, line 82: I wonder if we always make all template IDs available to both apps to avoid confusion/errors down the line?

apps/portal/src/app/views/applications/view/have-your-say/save.js
```diff
@@ -0,0 +1,181 @@
...
+async function populateNotificationData(req, service, answers, reference) {
+	const { db } = service;
+
+	const id = req.params.applicationId;
+	const crownDevelopment = await fetchPublishedApplication({
+		id,
+		db,
+		args: {
+			where: { id },
+			select: { reference: true, description: true, SiteAddress: true }
```
beejjacobs, line 110: Does this work? I thought relations had to be in the `include` field

packages/lib/forms/representations/save.js
beejjacobs, line 1: Perhaps it needs splitting out or refactoring - but its a shame to copy all this code into the two apps

beejjacobs (commented):
Generally looks pretty good - I am glad the service came in handy here!

A few comments to consider.

apps/portal/src/app/views/applications/view/have-your-say/save.js
tomxelliott: yeah it works but if include is the better practice I will use that instead.

beejjacobs: Checking the docs I misunderstood, you can do exactly this so leave it in!

packages/lib/forms/representations/save.js
tomxelliott: yeah have managed to refactor so that we can keep the common save function in lib

beejjacobs (approved):

PR #291, opened 2025-03-31
feat(manage): AUTOMATIC NOTIFICATION: To LPA - Acknowledge Receipt of Questionnaire

tomxelliott:
## Describe your changes
AUTOMATIC NOTIFICATION: To LPA - Acknowledge Receipt of Questionnaire
Amend copy on 'edit LPA questionnaire received date' page 

https://pins-ds.atlassian.net/browse/CROWN-262
https://pins-ds.atlassian.net/browse/CROWN-264


apps/manage/src/app/config-types.d.ts
```diff
@@ -44,4 +44,5 @@ interface Config {
 	};
 	staticDir: string;
 	govNotify: NotifyConfig;
+	frontOfficeLink: string;
```
beejjacobs, line 47: I think we should use `portal` here for consistency. Also it'd be similar if the env var had the protocol in it.

How about: `portalBaseUrl` or similar? Then in infra
`PORTAL_BASE_URL               = "https://${var.web_domains.portal}"`

We might have situations where we need to link to another part of the app, so it'd be good if the config variable was more generic.

(Also looks like we've missed a few, but this config object was meant to be largely sorted a-z)

apps/manage/src/app/config.js
```diff
@@ -119,9 +122,11 @@ export function loadConfig() {
 	}
 
 	const protocol = APP_HOSTNAME?.startsWith('localhost') ? 'http://' : 'https://';
+	const frontOfficeProtocol = PORTAL_HOSTNAME?.startsWith('localhost') ? 'http://' : 'https://';
 
 	config = {
 		appHostname: APP_HOSTNAME,
+		frontOfficeLink: `${frontOfficeProtocol}${PORTAL_HOSTNAME}/applications`,
```
beejjacobs, line 129: See comment on the types related to this.

packages/dynamic-forms/src/components/date/index.njk
```diff
@@ -75,6 +75,8 @@
                     ]
                 }) }}
 
+                {{ beforeSubmitButton }}
```
beejjacobs, line 78: Not sure if this works if its undefined, or if it needs

```
{% if beforeSubmitButton %}
  {{ beforeSubmitButton }}
{% endif %}
```

packages/dynamic-forms/src/components/date/question.js
beejjacobs, line 1: This is too specific to make a change for in the package - I need to have a think about how we might tackle this more generically.

beejjacobs (commented):
Generally looks good - lets chat in Teams about the warning message change.

packages/dynamic-forms/src/components/date/index.njk
tomxelliott: Does not result in an error currently but in the interest of defensive programming I will alter 👍

beejjacobs (approved):

PR #338, opened 2025-04-28
fix(portal): set cookie domain

beejjacobs:
## Describe your changes
Don't use the default Google cookie settings which uses the root domain. This conflicts with the NSIP site.

https://pins-ds.atlassian.net/browse/CROWN-636

### DIFF
--- a/apps/portal/src/util/cookies.test.js
+++ b/apps/portal/src/util/cookies.test.js
@@ -27,7 +27,7 @@ describe('cookies', () => {
 		it('should set res.locals.analyticsCookiesEnabled to true if the cookie is present and true', () => {
 			const req = { headers: { cookie: 'CrownDevAnalyticsEnabled=true' } };
 			const res = { locals: {} };
-			const middleware = buildAnalyticsCookiesMiddleware();
+			const middleware = buildAnalyticsCookiesMiddleware({});
 			middleware(req, res, mock.fn());
 			assert.strictEqual(res.locals.analyticsCookiesEnabled, true);
 			assert.strictEqual(res.locals.analyticsCookiesPreferenceSet, true);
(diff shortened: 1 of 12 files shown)


apps/portal/src/util/cookies.test.js
```diff
@@ -116,38 +116,41 @@ describe('cookies', () => {
...
+			assert.strictEqual(res.clearCookie.mock.callCount(), 2);
+			assert.strictEqual(res.clearCookie.mock.calls[0].arguments[0], '_ga');
+			assert.strictEqual(res.clearCookie.mock.calls[1].arguments[0], '_ga');
+			assert.strictEqual(res.clearCookie.mock.calls[1].arguments[1].domain, '.example.com');
 		});
 		it('should not clear any other cookies', () => {
 			const req = { headers: { cookie: 'name=value; name2=value2' } };
 			const res = { clearCookie: mock.fn() };
-			removeAnalyticsCookies(req, res);
+			removeAnalyticsCookies(req, res, 'example.com');
```
AlexMoulin, line 153: Is it worth adding some tests to check that it: 

Doesn't pickup cookie preferences from another site with the same domain
It doesn't override the cookie preference of another site with the same domain?

AlexMoulin (changes requested):
Quick question around tests

apps/portal/src/util/cookies.test.js
beejjacobs: >Doesn't pickup cookie preferences from another site with the same domain
This is browser behaviour that I don't think we can control.

>It doesn't override the cookie preference of another site with the same domain?
We can only set/control cookies on the app domain and higher-level domains. We are only setting them on the app domain.

There is a test to check that the domain is including in 'clearCookie' call, which I think covers this otherwise?

AlexMoulin (approved):

tomxelliott (approved):

PR #368, opened 2025-05-22
feat(manage): add validation before reviewing representation

AlexMoulin:
## Describe your changes
Add validation to manage representation
https://pins-ds.atlassian.net/browse/CROWN-417

### DIFF
--- a/apps/manage/src/app/views/cases/view/manage-reps/validation-middleware.js
+++ b/apps/manage/src/app/views/cases/view/manage-reps/validation-middleware.js
@@ -1,0 +1,229 @@
+import { REPRESENTATION_SUBMITTED_FOR_ID, REPRESENTED_TYPE_ID } from '@pins/crowndev-database/src/seed/data-static.js';
+import { notFoundHandler } from '@pins/crowndev-lib/middleware/errors.js';
+import { addSessionData } from '@pins/crowndev-lib/util/session.js';
+
+/**
+ * Validate a representation before it can be accepted, rejected or redacted
+ * @param {import('#service').ManageService} services
+ * @returns {import('express').Handler}
+ */
+export function buildValidateRepresentationMiddleware(services) {
+	return async (req, res, next) => {
+		const { db, logger } = services;
+		const representationRef = req.params.representationRef;
+		const id = req.params.id;
+
+		if (!id) {
+			throw new Error(`id param required`);
+		}
+		if (!representationRef) {
+			throw new Error(`representationRef param required`);
+		}
+		logger.info({ representationReference: representationRef }, 'validate representation');
+		const representation = await db.representation.findUnique({
+			where: { reference: representationRef },
+			include: {
+				SubmittedByContact: { include: { jobTitleOrRole: true, orgName: true } },
+				RepresentedContact: { include: { firstName: true, lastName: true } }
+			}
+		});
+
+		if (!representation) {
+			return notFoundHandler(req, res);
+		}
+
+		const originUrl = `/cases/${id}/manage-representations/${representationRef}`;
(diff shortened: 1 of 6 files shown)


apps/manage/src/app/views/cases/view/manage-reps/validation-middleware.js
```diff
@@ -0,0 +1,239 @@
...
+ * Validate a representation before it can be accepted, rejected or redacted
+ * @param {import('#service').ManageService} services
+ * @returns {import('express').Handler}
+ */
+export function buildValidateRepresentationMiddleware(services) {
+	return async (req, res, next) => {
+		const { db, logger } = services;
+		const representationRef = req.params.representationRef;
+		const id = req.params.id;
+		clearSessionData(req, representationRef, 'errors', 'representations');
```
tomxelliott, line 15: is this required in two places? I can see it also clears in the controller. Are both clearSessionData calls required - can one of them be removed?

apps/manage/src/app/views/cases/view/manage-reps/validation-middleware.js
```diff
@@ -0,0 +1,239 @@
...
+		if (!representation) {
+			return notFoundHandler(req, res);
+		}
+
+		const originUrl = `/cases/${id}/manage-representations/${representationRef}`;
+		const requiredAnswers = getRequiredAnswers(representation, originUrl);
+
+		const errors = [];
+		for (const requiredAnswer of requiredAnswers) {
+			if (typeof requiredAnswer.value === 'undefined' || requiredAnswer.value === '' || requiredAnswer.value === null) {
```
tomxelliott, line 40: I believe this can be simplified to:
`if (!requiredAnswer.value) {}`
since !requiredAnswer.value checks for: undefined, null, '', false, 0, or NaN

AlexMoulin: Some values in the database are booleans, false is a valid value. makeRequiredAnswer generates the list of required fields, the loop checks if they are valid (ie. not empty strings, not undefined, not null).

apps/manage/src/app/views/cases/view/manage-reps/validation-middleware.js
```diff
@@ -0,0 +1,239 @@
...
+		const representationRef = req.params.representationRef;
+		const id = req.params.id;
+		clearSessionData(req, representationRef, 'errors', 'representations');
+		if (!id) {
+			throw new Error(`id param required`);
+		}
+		if (!representationRef) {
+			throw new Error(`representationRef param required`);
+		}
+		logger.info({ representationReference: representationRef }, 'validate representation');
```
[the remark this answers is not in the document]
AlexMoulin: I put it above the findUnique in the database as it would contextualise a 404 if the database doesn't return anything. Can move it if you still

apps/manage/src/app/views/cases/view/manage-reps/validation-middleware.js
AlexMoulin: Quite right, will sort that out.

apps/manage/src/app/views/cases/view/manage-reps/validation-middleware.js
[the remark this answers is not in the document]
tomxelliott: no worries

apps/manage/src/app/views/cases/view/manage-reps/validation-middleware.js
tomxelliott: ok - can the if statement be simplified to just !required.value since that checks all of the falsy you are checking

tomxelliott (approved):

PR #447, opened 2025-07-01
fix(manage): hint now displaying under what is your address

RobBroadhead30:
## Describe your changes

the hint for page add-representation/myself then option post selected has been added under the h1 title "We will not publish your address"

image

438 https://pins-ds.atlassian.net/browse/CROWN-438

### DIFF
--- a/packages/dynamic-forms/src/components/address/question.test.js
+++ b/packages/dynamic-forms/src/components/address/question.test.js
@@ -124,12 +124,13 @@ describe('AddressQuestion', () => {
 
 		it('should include the hint under the h1 title', async () => {
 			const { question } = setup();
-			question.hint = 'We will not publish your address';
+			const hintText = 'We will not publish your address';
+			question.hint = hintText;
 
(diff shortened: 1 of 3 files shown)


packages/dynamic-forms/src/components/address/question.test.js
```diff
@@ -121,6 +121,17 @@ describe('AddressQuestion', () => {
...
 		});
+
+		it('should include the hint under the h1 title', async () => {
+			const { question } = setup();
+			question.hint = 'We will not publish your address';
+
+			const response = { journeyId: 'testJourney', answers: { siteAddress: testAddress } };
+			const model = await question.prepQuestionForRendering({}, { getBackLink: mock.fn(), response });
+
+			assert.strictEqual(model.question.hint, question.hint);
```
tomxelliott, line 132: It might be worth being a bit more explicit here. Say assigning the text 'We will not publish your address' to a constant variable, assigning question.hint to the variable and then using the constant in the strict equal comparison.

PR #449, opened 2025-07-01
fix(package): handle folder not found correctly when saving a representation with no attachments

AlexMoulin:
## Describe your changes
Fix a bug introduced when deleting folders when saving a representation, if no folders existed (ie no attachments were uploaded) it would throw. It now handles correctly.


RobBroadhead30 (approved):
Looks good to me

PR #453, opened 2025-07-02
feat(manage): allow users to add attachments to existing representations

AlexMoulin:
## Describe your changes
Allow users to add documents to existing representation in review and view.
https://pins-ds.atlassian.net/browse/CROWN-701
https://pins-ds.atlassian.net/browse/CROWN-709
https://pins-ds.atlassian.net/browse/CROWN-712
https://pins-ds.atlassian.net/browse/CROWN-702

### DIFF
--- a/apps/manage/src/app/views/cases/view/manage-reps/edit/controller.js
+++ b/apps/manage/src/app/views/cases/view/manage-reps/edit/controller.js
@@ -4,15 +4,27 @@ import { validateParams } from '../view/controller.js';
 import { addSessionData, clearSessionData, readSessionData } from '@pins/crowndev-lib/util/session.js';
 import { publishedRepresentationsAttachmentsRootFolderPath } from '@pins/crowndev-lib/util/sharepoint-path.js';
 import { BOOLEAN_OPTIONS } from '@planning-inspectorate/dynamic-forms/src/components/boolean/question.js';
+import {
+	deleteRepresentationAttachmentsFolder,
+	moveAttachmentsToCaseFolder
+} from '@pins/crowndev-lib/util/handle-attachments.js';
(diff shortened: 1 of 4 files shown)


apps/manage/src/app/views/cases/view/manage-reps/edit/controller.js
```diff
@@ -31,6 +42,61 @@ export function buildUpdateRepresentation({ db, logger, getSharePointDrive }) {
...
+			} catch (err) {
+				wrapPrismaError({
+					error: err,
+					logger,
+					message: 'adding representation attachments',
+					logParams: { id, representationRef }
+				});
+			}
+			try {
+				await deleteRepresentationAttachmentsFolderFn(
```
tomxelliott, line 82: If we delete the folder at this point there is no risk of it causing an issue with any other journeys that also might have folders in that particular session id folder in SharePoint? Just wanted to double check since you are deleting everything from sessionId level and beyond.

This will be deleting the nested folders such as applicationId/journeyId/myself (in add a rep or have your say journey) or applicationId/journeyId/itemId (in review and manage reps) so just want to make sure that is ok as they may still have an active session in review/manage reps. Have your say and add a rep is less of an issue since if they abandon half way through we have our session clearing mechanism if they re-enter the normal way, however if they re-enter by manually entering the link then they may lose their document session data. Just something to be aware of I suppose.

Apologies for the longer comment.

AlexMoulin: With how we are handling the files now this should not be a problem. There is no reason that they should be doing two journeys at once (especially with your change to create a new session for new journeys)

PR #462, opened 2025-07-08
feat(manage): add site visit field

tomxelliott:
## Describe your changes
Site visit field

https://pins-ds.atlassian.net/browse/CROWN-455

Screenshot 2025-07-08 at 17 32 31


apps/manage/src/app/sass/style.scss
```diff
@@ -26,3 +26,8 @@ $font-family: arial, helvetica, sans-serif;
 .status-tag-width {
 	max-width: 250px !important;
 }
+
+.govuk-select--width-4 {
```
AlexMoulin, line 30: I would avoid using .govuk as a class. Those should be reserved for the govuk package. You can use pins instead

AlexMoulin (approved):
LGTM

PR #475, opened 2025-07-18
feat(manage): withdraw representation

tomxelliott:
## Describe your changes
Withdraw representation journey

https://pins-ds.atlassian.net/browse/CROWN-583
https://pins-ds.atlassian.net/browse/CROWN-584
https://pins-ds.atlassian.net/browse/CROWN-585
https://pins-ds.atlassian.net/browse/CROWN-586
https://pins-ds.atlassian.net/browse/CROWN-587
https://pins-ds.atlassian.net/browse/CROWN-588

### DIFF
--- a/apps/manage/src/app/views/cases/view/manage-reps/withdraw/controller.js
+++ b/apps/manage/src/app/views/cases/view/manage-reps/withdraw/controller.js
@@ -1,0 +1,108 @@
+import { clearDataFromSession } from '@planning-inspectorate/dynamic-forms/src/lib/session-answer-store.js';
+import { JOURNEY_ID } from './journey.js';
+import { validateParams } from '../view/controller.js';
+import { wrapPrismaError } from '@pins/crowndev-lib/util/database.js';
+import { REPRESENTATION_STATUS_ID } from '@pins/crowndev-database/src/seed/data-static.js';
+import {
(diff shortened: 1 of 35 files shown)


apps/manage/src/app/views/cases/view/manage-reps/withdraw/controller.js
```diff
@@ -0,0 +1,108 @@
...
+				logger,
+				message: 'withdrawing representation',
+				logParams: { id }
+			});
+		}
+
+		const crownDevelopment = await db.crownDevelopment.findUnique({
+			where: { id },
+			select: { reference: true }
+		});
```
AlexMoulin, line 63: Can this be added to a transaction with the getUnique representation to reduce the number of DB queries?

AlexMoulin (changes requested):
Looking good overall. Some minor tweaks and it should be good to go.

apps/manage/src/app/views/cases/view/manage-reps/withdraw/controller.js
tomxelliott: since we use the crown development returned outside of the transaction for subsequent calls to sharepoint and it is a get/find rather than update/create I will leave it there if that is ok.

PR #487, opened 2025-07-24
feat(manage): add duration fields for hearing and inquiry prep sitting and reporting

RobBroadhead30:
## Describe your changes

update to hearing and inquiry durations from radio button to labels which accept numerical answers for prep, sitting and reporting fields and display under duration under manage case

hearing ticket - https://pins-ds.atlassian.net/browse/CROWN-500
inquiry ticket - https://pins-ds.atlassian.net/browse/CROWN-776

Screenshot 2025-07-24 122302

Screenshot 2025-07-24 122315

Screenshot 2025-07-24 122222

### DIFF
--- a/apps/manage/src/app/views/cases/view/question-utils.test.js
+++ b/apps/manage/src/app/views/cases/view/question-utils.test.js
@@ -64,5 +64,64 @@ describe('question-utils', () => {
 				assert.ok(question.url.startsWith(prefixHyphens));
 			}
 		});
+		it('should format user answer with formatPrefix and formatJoinString in duration', () => {
+			const prefix = 'hearing';
+			const questions = eventQuestions(prefix);
+			const durationQuestion = questions[`${prefix}Duration`];
+			const userAnswer = {
+				hearingDurationPrep: '2',
+				hearingDurationSitting: '3',
+				hearingDurationReporting: '1'
+			};
+			const formatted = durationQuestion.inputFields
+				.map((field) => {
+					const value = userAnswer[field.fieldName];
+					return value ? `${field.formatPrefix}${value}${field.formatJoinString}` : '';
+				})
+				.join('');
+
+			assert.ok(formatted.includes('Prep: 2 days'));
+			assert.ok(formatted.includes('Sitting: 3 days'));
+			assert.ok(formatted.includes('Reporting: 1 days'));
+		});
+
+		it('should format decimal values correctly in duration fields', () => {
+			const prefix = 'hearing';
+			const questions = eventQuestions(prefix);
+			const durationQuestion = questions[`${prefix}Duration`];
+			const userAnswer = {
+				hearingDurationPrep: '2.5',
+				hearingDurationSitting: '3.75',
+				hearingDurationReporting: '1.0'
+			};
+			const formatted = durationQuestion.inputFields
+				.map((field) => {
+					const value = userAnswer[field.fieldName];
+					return value ? `${field.formatPrefix}${value}${field.formatJoinString}` : '';
+				})
+				.join('');
+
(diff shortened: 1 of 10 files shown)


packages/lib/util/numbers.js
```diff
@@ -37,3 +37,17 @@ export function bytesToUnit(bytes, decimalPoints = 1) {
...
+ * @returns {*|number|string|null}
+ * @param value
+ */
+export function parseNumberStringToNumber(value) {
+	if (value === '' || value === null || value === undefined) return null;
+	if (typeof value === 'string') {
+		const num = Number(value);
+		return isNaN(num) ? value : num;
+	}
+	return Array.isArray(value) ? value : value;
```
AlexMoulin, line 52: This turnery seems redundant: if it's an array return value otherwise return value

apps/manage/src/app/views/cases/view/question-utils.test.js
```diff
@@ -64,5 +64,38 @@ describe('question-utils', () => {
 				assert.ok(question.url.startsWith(prefixHyphens));
 			}
 		});
+		it('should format user answer with formatPrefix and formatJoinString in duration', () => {
```
AlexMoulin, line 67: Could you also add a test to make sure it formats decimals correctly too

packages/database/src/schema.prisma
```diff
@@ -180,6 +180,9 @@ model Event {
   id                           String             @id @default(dbgenerated("newid()")) @db.UniqueIdentifier
   date                         DateTime?
   duration                     String?
```
AlexMoulin, line 182: I think we can remove duration (do a specific migration just for that so it can be reverted if required)

packages/database/src/schema.prisma
```diff
@@ -180,6 +180,9 @@ model Event {
...
+  prepDuration                 Decimal?           @db.Decimal(6, 2)
```
AlexMoulin, line 183: Using @db.Decimal(6,2) is a great idea! Nice!

apps/manage/src/app/views/cases/view/view-model.js
```diff
@@ -188,6 +188,23 @@ export function editsToDatabaseUpdates(edits, viewModel) {
 
 	if (hasProcedure(viewModel.procedureId)) {
 		const eventUpdates = viewModelToEventUpdateInput(edits, viewModel.procedureId);
+
+		if (eventUpdates.eventUpdateInput) {
+			const prefix = eventPrefix(viewModel.procedureId);
```
AlexMoulin, line 193: This seems to be duplicating/overwriting the assignments done on ln 427-436. Is this needed or do you just need to move the logic here and replace the ones lover down?

AlexMoulin (changes requested):
Some changes needed but the logic is sound.

AlexMoulin (dismissed):

apps/manage/src/app/views/cases/view/questions.js
```diff
@@ -324,7 +324,14 @@ export function getQuestions(groupMembers = { caseOfficers: [], inspectors: [] }
 			question: 'When is the site visit?',
 			fieldName: 'siteVisitDate',
 			url: 'site-visit',
-			validators: [new DateTimeValidator('Site visit', 'Site visit date')]
+			validators: [
+				new DateTimeValidator(
+					'Site visit',
+					'Site visit date',
+					{ ensureFuture: false, ensurePast: false },
```
RobBroadhead30, line 331: This is just reverting back to a change that was made previously which i accidentally removed when fixing the commit issues.

AlexMoulin (approved):
LGTM

PR #488, opened 2025-07-24
fix(manage): file upload no longer persists between representations

AlexMoulin:
## Describe your changes
Representation attachments were being saved by applicationId meaning  they were carried accross reps. Additionally, it didn't clear session so just kept them in session meaning they would persist incorrectlt.
https://pins-ds.atlassian.net/browse/CROWN-944


RobBroadhead30 (approved):
looks good to me

PR #512, opened 2025-08-06
feat(manage): add fee amount to display with commas for app and notify

RobBroadhead30:
## Describe your changes
Formats fee value to two decimal places with commas when displaying fee amount in case and notify.
909 https://pins-ds.atlassian.net/browse/CROWN-909

image

### DIFF
--- a/apps/manage/src/app/views/cases/view/update-case.js
+++ b/apps/manage/src/app/views/cases/view/update-case.js
@@ -86,6 +86,10 @@ function addCaseUpdatedSession(req, id) {
  * @param {import('./types.js').CrownDevelopmentViewModel} fullViewModel
  */
 export async function customUpdateCaseActions(service, id, toSave, fullViewModel) {
+	if (toSave.applicationFee !== undefined) {
+		handleApplicationFee(toSave);
+	}
+
 	if (toSave.lpaQuestionnaireReceivedDate && fullViewModel.lpaQuestionnaireReceivedEmailSent !== BOOLEAN_OPTIONS.YES) {
 		await handleLpaQuestionnaireReceivedDateUpdate(service, id, toSave);
 	}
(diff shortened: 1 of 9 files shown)


apps/manage/src/app/views/cases/view/update-case.js
```diff
@@ -37,6 +37,10 @@ export function buildUpdateCase(service, clearAnswer = false) {
 
 		await customUpdateCaseActions(service, id, toSave, fullViewModel);
 
+		if (typeof toSave.applicationFee === 'string') {
+			toSave.applicationFee = Number(toSave.applicationFee.replace(/,/g, ''));
+		}
+
```
AlexMoulin, line 43: Can this be moved to the customUpdateCaseActions as handleApplicationFee please?

packages/lib/forms/custom-components/fee-amount/fee-amount-validator.test.js
```diff
@@ -96,6 +96,76 @@ describe('./lib/forms/custom-components/fee-amount/fee-amount-validator.js', ()
...
+				applicationFee: 'yes',
+				applicationFee_amount: ''
+			}
+		};
+		const errors = await _validationMappedErrors(req, question);
+		assert.strictEqual(Object.keys(errors).length, 1);
+		assert.strictEqual(errors.applicationFee_amount.msg, 'Application fee is required');
+	});
+
+	it('should return an error message if fee amount contains commas', async () => {
```
AlexMoulin, line 111: This seems to be specific to decimals where as the next test is for integers (would be helpful to specify in the test name)

packages/lib/forms/custom-components/fee-amount/fee-amount-validator.js
```diff
@@ -17,12 +17,23 @@ export default class FeeAmountValidator extends BaseValidator {
 				.withMessage('Application fee is required')
 				.bail()
 				.custom((value) => {
-					if (!/^\d+(\.\d+)?$/.test(value)) {
+					const sanitized = (value || '').replace(/,/g, '');
+					if (!sanitized) {
+						throw new Error('Application fee is required');
+					}
+					if (/,/.test(value)) {
```
AlexMoulin, line 24: I am not sure I understand why you are sanitising the value here? The value that is being submitted to the DB is the "value" not the sanitised value.

AlexMoulin (changes requested):

PR #551, opened 2025-08-22
chore(portal): notify callback logging improved

AlexMoulin:


apps/manage/src/app/notify/controller.js
```diff
@@ -29,79 +29,100 @@ export function buildNotifyCallbackController(service) {
...
+		Status: { connect: { id: notification.status } },
+		templateId: notification.template?.id ?? null,
+		templateVersion: notification.template?.version ?? null,
+		body: notification.body ?? null,
+		subject: notification.subject ?? null
+	};
+}
+
+// Determines and applies association logic (Contact/Lpa) or falls back to email
+async function resolveNotificationAssociations({ db, logger }, { reference, emailAddress, notificationId }) {
```
tomxelliott, line 75: is there a reason the parameter can't be in a single object?

PR #559, opened 2025-08-29
feat(manage): removes data assigned to procedure when user changes procedure type

RobBroadhead30:
## Describe your changes
When a user has selected either Written Representation, Hearing or Inquiry first and added answers to that part of the form, if they then go to change the procedure type for example from Hearing to Inquiry the data assigned to Hearing will be removed and Inquiry answers will be blank for the user to complete.

Hint also added into procedure to warn user of this.

152 https://pins-ds.atlassian.net/browse/CROWN-152

image
image
image

### DIFF
--- a/apps/manage/src/app/views/cases/view/update-case.js
+++ b/apps/manage/src/app/views/cases/view/update-case.js
@@ -7,6 +7,45 @@ import {
 import { editsToDatabaseUpdates } from './view-model.js';
 import { wrapPrismaError } from '@pins/crowndev-lib/util/database.js';
 
+export async function clearProcedureData({ db }, edits, viewModel, applicationId) {
+	if (!('procedureId' in edits)) {
+		return {};
+	}
+	const validProcedures = ['hearing', 'inquiry', 'writtenReps'];
+	const procedureId = edits.procedureId;
+	if (procedureId && validProcedures.includes(procedureId)) {
+		const crownDevelopment = await db.crownDevelopment.findUnique({
+			where: { id: applicationId },
+			select: { procedureId: true, eventId: true }
+		});
+		if (
+			crownDevelopment?.eventId &&
+			crownDevelopment.procedureId !== null &&
+			crownDevelopment.procedureId !== procedureId
+		) {
+			return {
+				procedureNotificationDate: null,
+				Event: {
+					update: {
+						data: {
+							date: null,
+							prepDuration: null,
+							sittingDuration: null,
+							reportingDuration: null,
+							venue: null,
+							notificationDate: null,
+							issuesReportPublishedDate: null,
+							statementsDate: null,
+							caseManagementConferenceDate: null,
+							proofsOfEvidenceDate: null
+						}
+					}
+				}
+			};
+		}
+	}
+	return {};
+}
 /**
  * @param {import('#service').ManageService} service
  * @param {boolean} [clearAnswer=false] - whether to clear the answer before saving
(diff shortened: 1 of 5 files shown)


apps/manage/src/app/views/cases/view/update-case.js
```diff
@@ -7,6 +7,63 @@ import {
...
+		prepDuration: null,
+		sittingDuration: null,
+		reportDuration: null,
+		notificationDate: null,
+		issuesReportPublishedDate: null,
+		statementsDate: null,
+		caseManagementConferenceDate: null,
+		proofOfEvidenceDate: null
+	};
+}
```
AlexMoulin, line 26: Not needed

apps/manage/src/app/views/cases/view/update-case.js
```diff
@@ -7,6 +7,63 @@ import {
...
+	const eventTypeChanged = 'eventTypeId' in edits && edits.eventTypeId !== viewModel.eventTypeId;
+	const procedureChanged = 'procedureId' in edits && edits.procedureId !== viewModel.procedureId;
+
+	if (!eventTypeChanged && !procedureChanged) {
+		return {};
+	}
+	let clearedFields = {};
+	if (eventTypeChanged) {
+		Object.assign(clearedFields, getClearedEventTypeFields(edits.eventTypeId));
+	}
```
AlexMoulin, line 61: Not needed

apps/manage/src/app/views/cases/view/update-case.js
```diff
@@ -7,6 +7,63 @@ import {
...
+			upsert: {
+				where: eventId ? { id: eventId } : undefined,
+				create: clearedEventFields,
+				update: clearedEventFields
+			}
+		}
+	};
+}
+export function clearEventFieldsIfChanged(edits, viewModel) {
+	const eventTypeChanged = 'eventTypeId' in edits && edits.eventTypeId !== viewModel.eventTypeId;
```
AlexMoulin, line 52: Remove eventTypeChanged

apps/manage/src/app/views/cases/view/update-case.js
```diff
@@ -7,6 +7,63 @@ import {
...
+	const procedureChanged = 'procedureId' in edits && edits.procedureId !== viewModel.procedureId;
```
AlexMoulin, line 53: Do a db query here to check if the procedureId has chance (instead of comparing it to the viewModel). View model is only different because the value is in an array which means that if you change it from "hearing" to "hearing (ie don't change but still save) it will delete the date (probably not what you want)

apps/manage/src/app/views/cases/view/update-case.test.js
```diff
@@ -784,5 +784,173 @@ describe('case details', () => {
 			assert.strictEqual(updateArg.where?.id, 'case1');
 			assert.strictEqual(updateArg.data.description, null);
 		});
+		it('should clear event fields if event type changed', async (context) => {
```
AlexMoulin, line 787: Add a test for when the value in toSave is the same as it was before (ie technically doesn't change) to make sure it doesn't delete data in that case

apps/manage/src/app/views/cases/view/view-model.js
AlexMoulin, line 1: revert changes to this file.

AlexMoulin (changes requested):

apps/manage/src/app/views/cases/view/view-model.test.js
AlexMoulin, line 1: Remove

package.json
```diff
@@ -66,7 +66,7 @@
     "nodemon": "^3.1.10",
     "prettier": "^3.6.2",
     "prettier-plugin-prisma": "^5.0.0",
-    "prisma": "^6.14.0",
+    "prisma": "^6.15.0",
```
tomxelliott, line 69: why the update to prisma?

apps/manage/src/app/views/cases/view/update-case.js
```diff
@@ -7,6 +7,45 @@ import {
...
+	const validProcedures = ['hearing', 'inquiry', 'writtenReps'];
+	const procedureId = edits.procedureId;
+	if (procedureId && validProcedures.includes(procedureId)) {
+		const crownDevelopment = await db.crownDevelopment.findUnique({
+			where: { id: applicationId },
+			select: { procedureId: true, eventId: true }
+		});
+		if (
+			crownDevelopment?.eventId &&
+			crownDevelopment.procedureId !== null &&
```
tomxelliott, line 23: can this not just be `crownDevelopment.procedureId &&` since if it is not null it will be a truthy value?

PR #592, opened 2025-09-24
fix(app): add line breaks to representations for improved readability

RobBroadhead30:
## Describe your changes
Added the pins-pre-wrap class to redact-confirmation and review-comment.njk to add line breaks when present in the comment for readability on both Portal and Manage app.

https://pins-ds.atlassian.net/browse/CROWN-1097

Screenshot 2025-09-25 091525
Screenshot 2025-09-25 091542
Screenshot 2025-09-25 091558
Screenshot 2025-09-25 092832
Screenshot 2025-09-25 092623
Screenshot 2025-09-25 092644
Screenshot 2025-09-25 092704

### DIFF
--- a/apps/manage/src/app/views/cases/view/manage-reps/review/redact-confirmation.njk
+++ b/apps/manage/src/app/views/cases/view/manage-reps/review/redact-confirmation.njk
@@ -21,16 +21,17 @@
     </div>
 
     <div class="govuk-grid-row">
-        <div class="govuk-grid-column-full pins-pre-wrap">
+        <div class="govuk-grid-column-full">
             {{ govukSummaryList({
                 rows: [
                     {
                         key: {text: 'Original representation'},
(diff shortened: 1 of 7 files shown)


apps/manage/src/app/views/cases/view/manage-reps/review/redact-confirmation.njk
```diff
@@ -21,7 +21,7 @@
     </div>
 
     <div class="govuk-grid-row">
-        <div class="govuk-grid-column-full">
+        <div class="govuk-grid-column-full pins-pre-wrap">
```
tomxelliott, line 24: use the 'classes' attribute instead I think for this one like you did in the other file. By setting it on the whole div I think it is causing the unexpected behaviour so follow the approach you used in review-comment.njk

PR #607, opened 2025-10-01
fix(portal): fix test where date needs to be mocked due to a time window

AlexMoulin:
## Describe your changes
Fixed an issue where the timewindow to display a link was passed and the date was not mocked which caused it to fail


AlexMoulin:
Duplicate of fix in https://github.com/Planning-Inspectorate/crown-developments/pull/602/files#diff-b9e961487ba0be9c37fc1cf84f3753073abb0cd01d45ad3fb0f90f6979df9049L5

PR #612, opened 2025-10-03
fix(manage): fix an issue where removing a procedure doesn't work due to move to connection

AlexMoulin:
## Describe your changes
Removing procedure didn't work due to change of way we handle procedure update in editsToDatabase.


RobBroadhead30 (approved):
LGTM

PR #905, opened 2026-04-16
feat(manage): add application category question and validator

RobBroadhead30:
## Describe your changes

This pull request refactors how certain application questions are presented and handled in the case management journey. It consolidates three related boolean questions ("Environmental Impact Assessment", "Development plan", and "Right of way") into a single "Application category" multi-field input, updates validation and summary logic, and improves error handling and test coverage for these changes.

Key changes include:

**Journey and Question Refactoring:**
- Replaces separate questions (`environmentalImpactAssessment`, `developmentPlan`, `rightOfWay`) with a combined `applicationCategory` custom multi-field input in both the original and v2 journey flows (`createJourney`, `createJourneyV2`) in `journey.js`. [[1]](diffhunk://#diff-846f8614d75c5978efb2c77c658b68d7250d1a762c4bdb4a13a5b1f661e56461L96-R96) [[2]](diffhunk://#diff-846f8614d75c5978efb2c77c658b68d7250d1a762c4bdb4a13a5b1f661e56461L276-R274)
- Updates the question definitions to group the three boolean fields under the new `applicationCategory` multi-field input, with appropriate hints, options, and a new validator. [[1]](diffhunk://#diff-7d382a2eabb13f962be080134708fb61ec15a271a20e00436d55d110b9aa1f13R39) [[2]](diffhunk://#diff-7d382a2eabb13f962be080134708fb61ec15a271a20e00436d55d110b9aa1f13L883-R925)

**Form Component and Validation Enhancements:**
[...]

PR #907, opened 2026-04-20
feat(lib): allow CrossQuestionValidator to use body values instead of session answers

pinssbrett:
## Describe your changes
`CrossQuestionValidator` as built is designed to compare values between different answers in the session. However, if used on a multi-field input question, these values have not yet been saved to the session and are in the express body (like other standard validation behaviour). I've added the `useBodyValues` toggle to this validator to allow it to be used in that situation as well.
Also tests for the above, and explicitly specifying class properties.

Note that usage of `test` has already been corrected to `it` for other tests in https://github.com/Planning-Inspectorate/crown-developments/pull/905, so I've not addressed that here.

N/A


RobBroadhead30 (approved):
lgtm

AlexMoulin (approved):

PR #911, opened 2026-04-22
Feat/add typescript

pinssbrett:
## Describe your changes
Add typescript to the project.
1. Add `tsconfig` based on [that in the template service](https://github.com/Planning-Inspectorate/template-service/blob/main/tsconfig.json).
2. Extend `eslint.config.js` to handle both JavaScript and TypeScript. Based on that in the template service, but uses `recommended-type-checked` rules instead of just `recommended`, for detection of deeper issues.
3. Fix type errors in existing `*.d.ts` files

https://pins-ds.atlassian.net/browse/CROWN-1590


AlexMoulin (approved):
LGTM!

PR #917, opened 2026-04-24
Fix/crown 1593 delete address when organisation is deleted

pinssbrett:
## Describe your changes

1. Make sure that orphaned addresses are deleted when an organisation is deleted
2. Add tests for the above

https://pins-ds.atlassian.net/browse/CROWN-1593


AlexMoulin (approved):
LGTM

PR #938, opened 2026-05-05
fix(portal): fix beta banner feedback link not opening in new tab

AlexMoulin:
## Describe your changes
This pull request makes a minor update to the feedback link in the `banner.njk` template. The link to the feedback page now opens in a new tab.
https://pins-ds.atlassian.net/browse/CROWN-1500

### DIFF
--- a/apps/manage/src/app/views/cases/view/organisation-contact-updates.ts
+++ b/apps/manage/src/app/views/cases/view/organisation-contact-updates.ts
@@ -1,6 +1,6 @@
 import { ORGANISATION_ROLES_ID } from '@pins/crowndev-database/src/seed/data-static.js';
 import { optionalWhere } from '@pins/crowndev-lib/util/database.js';
-import { viewModelToAddressUpdateInput, isAddress, isSameAddress } from '@pins/crowndev-lib/util/address.js';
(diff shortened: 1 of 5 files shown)


apps/portal/src/app/views/layouts/components/core/banner.njk
```diff
@@ -4,5 +4,5 @@
   tag: {
     text: "Beta"
   },
-  html: 'This is a beta service - your <a class="govuk-link" href="' + config.serviceFeedbackUrl + '"> feedback</a> will help us to improve it'
+  html: 'This is a beta service - your <a class="govuk-link" href="' + config.serviceFeedbackUrl + '" target="_blank"> feedback</a> will help us to improve it'
```
pinssbrett: best practice when using `target="_blank"` is to [also use `rel="noreferrer"` for security](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#security_and_privacy)
```suggestion
  html: 'This is a beta service - your <a class="govuk-link" href="' + config.serviceFeedbackUrl + '" target="_blank" rel="noreferrer"> feedback</a> will help us to improve it'
```

pinssbrett (changes requested):

PR #943, opened 2026-05-06
feat(manage): implement site address sharing for linked cases updates

RobBroadhead30:
## Describe your changes
This pull request introduces support for sharing and updating a site address across multiple linked cases. The main changes involve extracting and handling shared site address updates when multiple cases are being updated together, ensuring that site address changes are properly created or updated and linked to all relevant cases. The changes also include comprehensive tests to verify this new behavior.

Key changes include:

**Site Address Sharing Logic:**

* Enhanced `buildCaseUpdateWritePlan` to extract site address updates from the scalar input when multiple cases are involved, and to distinguish between updating an existing address and creating a new shared address. The plan now includes `siteAddressUpdate` and `siteAddressCreateData` fields for this purpose. [[1]](diffhunk://#diff-9932e4664e385c5d302837758db022cb38e9e8ec88a81916e54c0c64f9c09808R19-R20) [[2]](diffhunk://#diff-9932e4664e385c5d302837758db022cb38e9e8ec88a81916e54c0c64f9c09808L94-R104) [[3]](diffhunk://#diff-9932e4664e385c5d302837758db022cb38e9e8ec88a81916e54c0c64f9c09808R119-R137)
* Updated `executeCaseUpdateWritePlan` to handle the new plan fields: it updates an existing address or creates a new address as needed, and ensures all relevant cases are linked to the shared address. [[1]](diffhunk://#diff-9932e4664e385c5d302837758db022cb38e9e8ec88a81916e54c0c64f9c09808R422-R435) [[2]](diffhunk://#diff-9932e4664e385c5d302837758db022cb38e9e8ec88a81916e54c0c64f9c09808L437-R481)
* Modified the update-case logic to determine the correct shared site address ID among linked cases, checking parent and child cases for an existing address, and passing this information to the plan builder. [[1]](diffhunk://#diff-550b9c354792f18e6f8531263642e52f6484b912048e9c3fc9a2e4e0e5fd60a0L55-R56) [[2]](diffhunk://#diff-550b9c354792f18e6f8531263642e52f6484b912048e9c3fc9a2e4e0e5fd60a0R112-R135)

**Testing:**

* Added new tests covering scenarios for extracting and applying site address updates when updating multiple linked cases, ensuring correct creation, updating, and linking of shared addresses, as well as verifying that single-case updates do not trigger shared address logic.
https://pins-ds.atlassian.net/browse/CROWN-1600
image
image
image

### DIFF
--- a/apps/manage/src/app/views/cases/view/organisation-contact-updates.ts
+++ b/apps/manage/src/app/views/cases/view/organisation-contact-updates.ts
@@ -16,6 +16,8 @@ import type { ManageService } from '#service';
 
 type CaseUpdateWritePlan = {
 	scalarCaseUpdates: { caseIds: string[]; updateInput: Prisma.CrownDevelopmentUpdateInput };
+	siteAddressUpdate?: { addressId: string | null; addressData: Prisma.AddressUpdateInput } | null;
+	siteAddressCreateData?: Prisma.AddressCreateInput;
 	shared: {
 		organisationCreates: Array<{ placeholderId: string; data: Prisma.OrganisationCreateInput }>;
 		organisationUpdates: Array<{ organisationId: string; data: Prisma.OrganisationUpdateInput }>;
@@ -91,13 +93,15 @@ export function buildCaseUpdateWritePlan({
 	dbViewModel,
 	caseIds,
 	scalarUpdateInput,
-	crownDevelopments
+	crownDevelopments,
+	sharedSiteAddressId
 }: {
 	toSave: Partial<CrownDevelopmentViewModel>;
 	dbViewModel: CrownDevelopmentViewModel;
 	caseIds: string[];
 	scalarUpdateInput: Prisma.CrownDevelopmentUpdateInput;
 	crownDevelopments: CrownDevelopmentPayload[];
+	sharedSiteAddressId?: string | null;
 }): CaseUpdateWritePlan {
 	const plan: CaseUpdateWritePlan = {
 		scalarCaseUpdates: { caseIds, updateInput: scalarUpdateInput },
@@ -111,6 +115,25 @@ export function buildCaseUpdateWritePlan({
 			organisationToContactCreates: []
 		}
 	};
+
+	if (caseIds.length > 1 && scalarUpdateInput.SiteAddress && 'upsert' in scalarUpdateInput.SiteAddress) {
+		const upsert = scalarUpdateInput.SiteAddress.upsert;
+		if (upsert) {
+			const addressData = {
+				create: upsert.create as Prisma.AddressCreateInput,
+				update: upsert.update as Prisma.AddressUpdateInput
+			};
+
+			plan.siteAddressUpdate = {
+				addressId: sharedSiteAddressId ?? null,
+				addressData: addressData.update
+			};
+
+			plan.siteAddressCreateData = addressData.create;
+
+			delete scalarUpdateInput.SiteAddress;
+		}
+	}
 
 	const existingLinks = new Set();
 	const organisationById: Map<string, Prisma.OrganisationGetPayload<object>> = new Map();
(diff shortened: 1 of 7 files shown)


apps/manage/src/app/views/cases/view/organisation-contact-updates.ts
```diff
@@ -16,6 +16,8 @@ import type { ManageService } from '#service';
 
 type CaseUpdateWritePlan = {
 	scalarCaseUpdates: { caseIds: string[]; updateInput: Prisma.CrownDevelopmentUpdateInput };
+	siteAddressUpdate?: { addressId: string | null; addressData: Prisma.AddressUpdateInput } | null;
+	siteAddressCreateData?: Prisma.AddressCreateInput;
 	shared: {
```
pinssbrett: Now that this plan is no longer just updating `Organisations`, this needs renaming, because the addresses are shared too, just not part of this. Will require lots of renaming across the file!
```suggestion
	organisationGraph: {
```

apps/manage/src/app/views/cases/view/organisation-contact-updates.ts
```diff
@@ -112,6 +116,25 @@ export function buildCaseUpdateWritePlan({
 		}
 	};
 
+	if (caseIds.length > 1 && scalarUpdateInput.SiteAddress && 'upsert' in scalarUpdateInput.SiteAddress) {
+		const upsert = scalarUpdateInput.SiteAddress.upsert;
+		if (upsert) {
+			const addressData = {
+				create: upsert.create as Prisma.AddressCreateInput,
+				update: upsert.update as Prisma.AddressUpdateInput
```
pinssbrett: I think best to just let the types be inferred here: no benefit from casting them
```suggestion
				create: upsert.create,
				update: upsert.update
```

apps/manage/src/app/views/cases/view/organisation-contact-updates.ts
```diff
@@ -396,6 +419,20 @@ export async function executeCaseUpdateWritePlan(plan: CaseUpdateWritePlan, tx:
...
+	let sharedSiteAddressId: string | null = null;
+	if (plan.siteAddressUpdate) {
+		const { addressId, addressData } = plan.siteAddressUpdate;
+
+		if (addressId) {
+			await tx.address.update({ where: { id: addressId }, data: addressData });
+			sharedSiteAddressId = addressId;
+		} else {
+			const createData = plan.siteAddressCreateData || (addressData as unknown as Prisma.AddressCreateInput);
+			const created = await tx.address.create({ data: createData });
```
pinssbrett: The type of `addressData` here is actually `Prisma.AddressUpdateInput` but you are saying it's `Prisma.AddressCreateInput` instead, because that's what `address.create` expects. What actually needs to happen is to resolve this discrepancy, rather than passing the wrong type of thing to Prisma but telling TS to ignore it.

This is caused by the extra fallback to the `UpdateInput` if the `CreateInput` doesn't exist. But it should always exist since we are setting them in the same place. So can just add an error if it's not there when we expect it to be.

```suggestion
		} else {
			if (!plan.siteAddressCreateData) {
				throw new Error('Missing AddressCreateInput for shared site address creation');
			}
			const created = await tx.address.create({ data: plan.siteAddressCreateData });
```

apps/manage/src/app/views/cases/view/organisation-contact-updates.ts
pinssbrett, line 1: I would rename this file from `organisation-contact-updates` to `linked-case-updates`, since the scope is wider now

apps/manage/src/app/views/cases/view/update-case.test.js
```diff
@@ -307,6 +307,391 @@ describe('case details', () => {
...
+			assert.strictEqual(mockDb.address.update.mock.callCount(), 1);
+			const addressUpdateArg = mockDb.address.update.mock.calls[0].arguments[0];
+			assert.strictEqual(addressUpdateArg.where?.id, 'child-address-1');
+
+			// Both cases should be connected to the child's address
+			const parentUpdate = mockDb.crownDevelopment.update.mock.calls[0].arguments[0];
+			const childUpdate = mockDb.crownDevelopment.update.mock.calls[1].arguments[0];
+			assert.strictEqual(parentUpdate.data?.SiteAddress?.connect?.id, 'child-address-1');
+			assert.strictEqual(childUpdate.data?.SiteAddress?.connect?.id, 'child-address-1');
+		});
```
pinssbrett: I think this is the same test as line 453

pinssbrett (changes requested):
Great work on a complex problem. There are some type things that need changing for best practice, and I think some names need to be change because of how this work extends the scope of the linked updates. Nice one!

pinssbrett (approved):
All points addressed - looks good!

PR #946, opened 2026-05-07
refactor(lib): migrate linked-case utility to TypeScript and update imports

pinssbrett:
## Describe your changes
Migrate `linked-case` util to typescript 

N/A


packages/lib/util/linked-case.test.js
```diff
@@ -7,7 +7,7 @@ import {
 	getSummaryWarningMessage,
 	hasLinkedCase,
 	linkedCaseIsPublished
-} from './linked-case.js';
+} from './linked-case.ts';
 
 describe('linked case util', () => {
```
paymonsattar, line 12: Is it worth also converting the tests to TS?

AlexMoulin (dismissed):
LGTM

packages/lib/util/linked-case.test.js
pinssbrett: let's discuss as a team - I think we need to resolve some things around how we treat types in tests first. also lower priority than converting the code

AlexMoulin (approved):
LGTM

PR #969, opened 2026-05-21
feat(database): add app history table

paymonsattar:
## Describe your changes

Adds the application history table to the crown schema. Also adds a new `updatedById` field to the application table that allows us to keep a record as to who made the last update on that application

### DIFF
--- a/packages/database/src/schema.prisma
+++ b/packages/database/src/schema.prisma
@@ -26,6 +26,8 @@ model CrownDevelopment {
   reference                String                           @unique
   createdDate              DateTime                         @default(now())
   updatedDate              DateTime?
+  /// Entra ID
+  updatedById              String?
   /// type of application, such as outline, extension
   typeId                   String
   Type                     ApplicationType                  @relation(fields: [typeId], references: [id])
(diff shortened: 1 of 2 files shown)


packages/database/src/schema.prisma
```diff
@@ -178,6 +181,19 @@ model Address {
   Organisation     Organisation[]
 }
 
+model ApplicationHistory {
```
oscargay, line 185: Might be worth leaving a `/// ` comment explaining what this table is used for / going to be used for and why it's currently not actually being used or populated?

packages/database/src/schema.prisma
```diff
@@ -178,6 +181,19 @@ model Address {
...
+  id            String           @id @default(dbgenerated("newid()")) @db.UniqueIdentifier
+  applicationId String           @db.UniqueIdentifier
+  Application   CrownDevelopment @relation(fields: [applicationId], references: [id])
+  action        String
+  metadata      String?          @db.NVarChar(MAX)
+  /// Entra ID
+  userId        String?
```
oscargay: There are audits that can be created without a user right? Is that why `userId` is nullable?

oscargay (commented):
Looks good! 1 minor point about a comment, and then just a question for my understanding !

packages/database/src/schema.prisma
paymonsattar: Good point, I don't see a case for audits not requiring a userId

pinssbrett (dismissed):
Looks good!

oscargay (approved):
LGTM

PR #982, opened 2026-05-26
fix(database): add defaults to required value migration

pinssbrett:
## Describe your changes
Add default values for fields marked as required in migration in https://github.com/Planning-Inspectorate/crown-developments/pull/960
On the whole these fields should only be needed/used for seeded data on non-prod environments where the required fields are missing.

N/A


RobBroadhead30 (approved):
lgtm

oscargay (approved):
Looks good, the only thing obviously is that `lpaId` wasn't backfilled but I am guessing that that is because it's a foreign key and there is no realistic scenario in prod or seed data where we will have nothing here!

PR #983, opened 2026-05-26
feat(manage): add app history service

paymonsattar:
## Describe your changes
Introduces the audit service layer - the core capability for recording and retrieving case history (audit) events against a CrownDevelopment. This is the foundation for the case-history feature; it adds the ability to record and query audit events but does not yet wire it into any user-facing flow (the history page and event recording on case actions come in follow-up PRs).

- Audit service (audit/service.ts) — record, recordMany, getAllForCase, countForCase, and getLastModifiedInfo. record/recordMany are fire-and-forget: audit failures are logged but never thrown, so they can't break the user operation that triggered them. Reads fail soft (empty array / zero) so the history page can still render.
- Types (audit/types.ts) — AuditEntry (input contract, keyed on caseId), AuditEvent (the stored row shape, keyed on applicationId to match the schema), and AuditQueryOptions for pagination.
- Actions & templates (audit/actions.ts) — the action definitions and the resolveTemplate helper that turns an action + metadata into the human-readable "Details" string shown in the history table.
- Audit formatters (packages/lib/util/audit-formatters.ts) — display formatters used by the service, including formatDateTime for last-modified display.

### DIFF
--- a/apps/manage/src/app/audit/resolvers/README.md
+++ b/apps/manage/src/app/audit/resolvers/README.md
@@ -1,0 +1,33 @@
+# Audit Resolvers
+
+This directory is for **resolvers** that translate raw database and form data into human-readable audit entries.
+
+> **Status: not yet implemented.** This directory is currently empty. The audit
+> system today records only simple, self-describing events (e.g. `CASE_CREATED`)
+> whose metadata is supplied directly at the call site. Resolvers will be added
+> when the audit trail is extended to cover field-level and list changes. This
+> README describes the intended pattern for that work — it does not describe code
+> that exists yet.
+
+## Why resolvers will exist
+
+When a user saves a form, the raw data is database IDs, field keys, and values. An audit trail should read _"Status was updated from New to Accepted"_, not _"statusId was updated from new to accepted"_. A resolver knows how to translate one type of field from raw data into something a person can read.
+
+## Intended pattern
+
+### Field resolvers (scalar fields)
+
+For simple scalar fields on the `CrownDevelopment` model (status, procedure, case officer, etc.), a field resolver would take the old DB value and the new form value and return display-friendly `{ oldValue, newValue }` strings. Fields without a specific resolver would fall through to a default that stringifies the raw value.
+
+### List resolvers (one-to-many relationships)
+
+For one-to-many relationships, a list resolver would compare the old and new lists, determine what was added, removed, or changed, and produce the appropriate audit entries for each difference. Each entity type would get its own resolver.
+
+## When you add resolver support for a new field or entity
+
+1. **Add audit action(s)** in `audit/actions.ts` (e.g. `MY_ENTITY_ADDED`, `MY_ENTITY_UPDATED`, `MY_ENTITY_DELETED`).
+2. **Add matching templates** in `AUDIT_TEMPLATES` in the same file.
+3. **Create the resolver here** — a field resolver for a scalar field, or a new list resolver for a new entity type.
+4. **Wire it into the relevant save/update controller** so the resolver runs after the database write and the resulting entries are passed to `audit.recordMany()`.
+
+Update this README's status note once the first resolver lands.
--- a/apps/manage/src/app/audit/service.ts
+++ b/apps/manage/src/app/audit/service.ts
@@ -1,0 +1,219 @@
+import type { PrismaClient } from '@pins/crowndev-database/src/client/client.ts';
+import type { Logger } from 'pino';
+import { parseMetadata, type AuditEntry, type AuditEvent, type AuditQueryOptions } from './types.ts';
+import type { EntraGroupMembers } from '../../util/entra-groups.ts';
+import { formatDateTime } from '@pins/crowndev-lib/util/audit-formatters.ts';
+
+/**
+ * Builds the audit service used to record and retrieve case history events.
+ *
+ * Design notes:
+ *   - `record()` is fire-and-forget: audit failures are logged but never
+ *     thrown so they can't break the user's actual operation.
+ */
+export function buildAuditService(db: PrismaClient, logger: Logger) {
+	return {
+		/**
+		 * Persist a single audit event.
+		 *
+		 * Safe to call without awaiting if the caller doesn't need
+		 * confirmation, but awaiting is fine too
+		 */
+		async record(entry: AuditEntry): Promise<void> {
+			try {
+				if (!entry.userId) {
+					throw new Error('Cannot record audit event without a userId');
+				}
+
+				await db.$transaction([
+					db.applicationHistory.create({
+						data: {
+							Application: {
+								connect: { id: entry.caseId }
+							},
+							action: entry.action,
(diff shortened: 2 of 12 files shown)


apps/manage/src/app/audit/resolvers/README.md
```diff
@@ -0,0 +1,33 @@
...
+Each entity type has its own file:
+
+## Adding audit support for a new feature
+
+When you add a new field or entity type to the case model, you need to:
+
+1. **Add an audit action** in `audit/actions.ts` (e.g. `MY_ENTITY_ADDED`, `MY_ENTITY_UPDATED`, `MY_ENTITY_DELETED`)
+2. **Add a template** in `AUDIT_TEMPLATES` in the same file
+3. **Create or update a resolver** — either add to `FIELD_RESOLVERS` for a scalar field, or create a new list resolver file for a new entity type
+4. **Wire it up in `update-case.ts`** — add the resolver call in the audit section after the DB update
```
oscargay: Might be worth running through the `README`s to make sure there isn't any incorrect / MPESC specific info in it like referencing `update-case.ts`

apps/manage/src/app/audit/service.ts
```diff
@@ -0,0 +1,211 @@
...
+			try {
+				await db.$transaction([
+					db.applicationHistory.create({
+						data: {
+							Application: {
+								connect: { id: entry.caseId }
+							},
+							action: entry.action,
+							metadata: JSON.stringify(entry.metadata ?? {}),
+							userId: entry.userId || ''
```
oscargay: Is there an actual realistic scenario of `entry.userId` being falsy? If there is maybe we should treat it like an error before the `try-catch` (or even throwing an error right in the try-catch to log it below) rather than proceeding with the `insert` as it's probably telling of something going quite wrong?

apps/manage/src/app/audit/service.ts
```diff
@@ -0,0 +1,211 @@
...
+		 * crownDevelopment.update sets updatedDate/updatedById once at the end.
+		 */
+		async recordMany(entries: AuditEntry[]): Promise<void> {
+			if (entries.length === 0) return;
+
+			try {
+				const caseId = entries[0].caseId;
+				const userId = entries[0].userId;
+
+				await db.$transaction(async ($tx) => {
```
oscargay: I just realised, this transaction is interactive when it doesn't need to be as you don't use any logic or the results of query 1 in query 2, so it can be made the same as the one above (I think) which should provide a slight performance increase / a bit more dev consistency

apps/manage/src/app/audit/service.ts
```diff
@@ -0,0 +1,211 @@
...
+					}
+				});
+
+				if (!caseRow) {
+					throw new Error(`No case found for id: ${caseId}`);
+				}
+
+				const updatedDate = caseRow.updatedDate ? formatDateTime(caseRow.updatedDate) : null;
+
+				const allMembers = [...groupMembers.caseOfficers, ...groupMembers.inspectors];
```
oscargay, line 193: I'm about 90% sure that on Crown, unlike MPESC, `caseOfficers` group is just anyone who has access to the app rather than a special group, worth checking because if true you can just use that group rather than having to spread

apps/manage/src/app/audit/types.ts
```diff
@@ -0,0 +1,73 @@
...
+
+	return null;
+}
+
+/**
+ * Actions that support a collapsible file list in the details column.
+ * When these actions have a `files` array in metadata, the details
+ * will include a GOV.UK details component for show/hide.
+ */
+export const BULK_FILE_ACTIONS = new Set<string>([]);
```
oscargay: NIT: Potentially worth just removing this for now, and then adding it in when we add bulk file actions. Only because I have a feeling bulk file actions might be a long way away which will mean we will have this floating around for a while for no reason. What do you think? Unless there are bulk file actions we are planning to connect with sharepoint soon?

packages/lib/util/audit-formatters.ts
```diff
@@ -0,0 +1,155 @@
...
+	if (isNaN(date.getTime())) {
+		return '-';
+	}
+
+	return date.toLocaleDateString('en-GB', {
+		day: 'numeric',
+		month: 'long',
+		year: 'numeric',
+		timeZone: 'Europe/London'
+	});
```
oscargay: Thinking we might as well use `formatInTimeZone` for consistency actually here?

oscargay (commented):
Looks good, had a few questions which are mainly things that I think I missed in the original review and wanted to get your thoughts on. All small things!

apps/manage/src/app/audit/service.ts
paymonsattar: There isn't, I've just added this as user Id can't be null in the schema. Might look if theres a better way to this upstream

apps/manage/src/app/audit/service.ts
AlexMoulin: We don't currently use two groups but I would suggest sticking with the way Paymon has done it so that if/when we implement role based access it just works

apps/manage/src/app/audit/types.ts
paymonsattar: Removed it! Thanks

packages/lib/util/audit-formatters.ts
paymonsattar: Good shout

packages/lib/util/audit-formatters.ts
```diff
@@ -0,0 +1,150 @@
...
+	const timeInUK = formatInTimeZone(date, 'Europe/London', 'HH:mm');
+	if (timeInUK === '00:00') {
+		return formatInTimeZone(date, 'Europe/London', 'd MMMM yyyy');
+	}
+
+	return formatInTimeZone(date, 'Europe/London', 'd MMMM yyyy h:mmaaa');
+}
+
+/**
+ * Formats a date value for audit display.
```
oscargay: Apologies, again I appreciate this is ported over but actually I just realised these 2 functions are basically identical. I think it's probably just worth making them 1 function and adding a toggle for whether you want the time, that way there's one nice source of truth for your dates. What do you think? Or could do a helper function if you like having them separate but I think it makes sense to do 1


```suggestion
/**
 * Formats a date value for audit display, optionally including the time.
 */
export function formatDate(value: Date | string | null | undefined, includeTime = false): string {
	if (!value) {
		return '-';
	}

	const date = value instanceof Date ? value : new Date(value);

	if (isNaN(date.getTime())) {
		return '-';
	}

	if (!includeTime || formatInTimeZone(date, 'Europe/London', 'HH:mm') === '00:00') {
		return formatInTimeZone(date, 'Europe/London', 'd MMMM yyyy');
	}

	return formatInTimeZone(date, 'Europe/London', 'd MMMM yyyy h:mmaaa');
}
```

oscargay (commented):
Just 1 comment! Sorry I missed it before...

oscargay (approved):
lgtm!

PR #984, opened 2026-05-26
Feat/peas 6 app history implementation

paymonsattar:
## Describe your changes
Adds the user-facing read path for audit events: a paginated case history page showing all audit events for a case, and a last-modified summary on the case details page. Builds on the audit service foundation ([PR](https://github.com/Planning-Inspectorate/crown-developments/pull/983)).

Screenshot 2026-05-26 at 17 01 54
Screenshot 2026-05-26 at 17 02 11


apps/manage/src/app/views/cases/case-history/view-model.ts
```diff
@@ -0,0 +1,29 @@
+import { formatDateTime } from '@pins/crowndev-lib/util/audit-formatters.ts';
+import { resolveTemplate, type AuditAction } from '../../../audit/actions.ts';
+import type { AuditEvent } from '../../../audit/types.ts';
+
+export interface CaseHistoryRow {
+	/** Formatted date+time, e.g. "11 February 2026 2:31pm" */
+	dateTime: string;
```
pinssbrett: optional: might a different name help distinguish this from a `Date`?
```suggestion
	dateTimeFormatted: string;
```

apps/manage/src/app/views/cases/case-history/view.njk
```diff
@@ -0,0 +1,99 @@
...
+                {% else %}
+                    <p class="govuk-body govuk-clearfix">
+                        <p class="govuk-body" data-cy="pagination-results-text">
+                            Showing <strong>{{ paginationParams.resultsStartNumber }}</strong> to <strong>{{ paginationParams.resultsEndNumber if paginationParams.totalCount > paginationParams.resultsEndNumber else paginationParams.totalCount }}</strong>
+                            of <strong>{{ paginationParams.totalCount }}</strong> results
+                        </p>
+                        <p class="pins-pagination-options">
+                            {{ paginationOptions.renderPagination(paginationParams.selectedItemsPerPage, paginationParams.totalCount, baseUrl, queryParams) }}
+                        </p>
+                    </p>
```
pinssbrett, line 34: nested `p` elements are not valid HTML. pagination should also probably be in a `nav` element for screen readers.

package-lock.json
```diff
@@ -1538,9 +1539,9 @@
       "link": true
     },
     "node_modules/@planning-inspectorate/dynamic-forms": {
-      "version": "3.9.0",
-      "resolved": "https://registry.npmjs.org/@planning-inspectorate/dynamic-forms/-/dynamic-forms-3.9.0.tgz",
-      "integrity": "sha512-H25j+rE/V5ecpO5vWESGkzDPEHRD8iJwvCPWPdGlfE0HRxh6Snl/4Sz6KXpd1f99GhZxBiwqylj+MDeWpAjZ+w==",
+      "version": "3.10.2",
```
oscargay, line 1542: Do you know what else is being pulled across in the dynamic-forms version bump? Just to make sure no breaking changes or anything

apps/manage/src/app/views/cases/case-history/controller.ts
```diff
@@ -0,0 +1,92 @@
...
+
+		const eventsWithUserNames = events.map((event) => ({
+			...event,
+			userName: userMap.get(event.userId ?? '') ?? 'Unknown User'
+		}));
+
+		const rows = createCaseHistoryViewModel(eventsWithUserNames);
+
+		return res.render('views/cases/case-history/view.njk', {
+			pageHeading: 'Case history',
```
oscargay, line 82: Are we calling it `Case history` or `Application history` think I've seen both used so would be good for consistency

apps/manage/src/app/views/cases/case-history/controller.ts
```diff
@@ -0,0 +1,92 @@
...
+			reference: caseRow.reference,
+			backLinkUrl: `/cases/${id}`,
+			backLinkText: 'Back to case details',
```
oscargay, line 85: Like above, do they call it case details in crown, or is it like application page or something?

apps/manage/src/app/views/cases/view/index.js
```diff
@@ -47,6 +49,9 @@ export function createRoutes(service) {
 		router.use('/application-updates', applicationUpdates);
 	}
 
+	// View case history page, /:id/case-history
+	router.use('/case-history', caseHistoryRoutes);
```
oscargay, line 53: Again just wanna make sure we have alignment over what to call this, case vs application.

oscargay (commented):
Looks good! Just 1 comment (across 3) about naming, do we know if wanna call this application history or case?

apps/manage/src/app/views/cases/case-history/view-model.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+/**
+ * Transforms raw audit events into rows ready for the case history table.
+ */
+export function createCaseHistoryViewModel(events: Array<AuditEvent & { userName: string }>): CaseHistoryRow[] {
+	return events.map((event) => {
+		return {
+			dateTime: formatDateTime(new Date(event.createdAt)),
+			details: resolveTemplate(event.action as AuditAction, event.metadata ?? undefined),
```
pinssbrett: If `event.action` is an untrusted DB string, should we validate it as `AuditAction` instead of casting the type?

apps/manage/src/app/views/cases/view/view.njk
```diff
@@ -122,6 +122,49 @@
...
+                    },
+                    attributes: {id: "case-audit-log"}
+                },
+                rows: [
+                    {
+                        key: {
+                            text: "Last modified date"
+                        },
+                        value: {
+                            text: lastModifiedDate if lastModifiedDate else "-"
```
pinssbrett: Can this fallback logic be handled in the controller rather than in the template?

pinssbrett (changes requested):
Looks good! A few minor comments about types, names and HTML.

apps/manage/src/app/views/cases/case-history/controller.ts
paymonsattar: So I've found Crown a little weird, application is used in the DB layer as the table is called 'application' and then we seem to use 'case' in the FE... So I've tried to follow a similar thing here

apps/manage/src/app/views/cases/case-history/controller.ts
paymonsattar: I believe it's case details

apps/manage/src/app/views/cases/view/index.js
paymonsattar: Same as above

package-lock.json
paymonsattar: It's not a major version update so I don't think theres any breaking changes.. 

i've given it a test anyway and don't see anything broken

apps/manage/src/app/views/cases/case-history/view-model.ts
paymonsattar: I don't really see a need for this... `resolveTemplate` already handles unknown actions safely (it just returns undefined), and the only writer to this column is the typed audit service, so an invalid value can't realistically appear.

apps/manage/src/app/views/cases/view/view.njk
paymonsattar: Good shout - moved it to the controller

apps/manage/src/app/views/cases/case-history/view.njk
paymonsattar: Good point here, thanks.

apps/manage/src/app/views/cases/view/view.njk
```diff
@@ -97,6 +97,49 @@
                     }) }}
                 {% endif %}
             {% endfor %}
+            {{ govukSummaryList({
+                card: {
+                    title: {
+                        text: "Application audit log"
```
oscargay: Think this one probably needs to change too?

apps/manage/src/app/views/cases/view/controller.ts
```diff
@@ -114,6 +114,12 @@ export function buildViewCaseDetails({ db, getSharePointDrive, isApplicationUpda
 		const baseUrl = req.baseUrl;
 
 		const banner = await getBannerMessages(id, res, req, db);
+		const lastModified = res.locals.lastModified as { updatedDate: string | null; by: string | null } | undefined;
+		const lastModifiedDate = lastModified?.updatedDate ?? '-';
+		const lastModifiedBy = lastModified?.by ?? '-';
+		const successParam = req.query.success;
+		const casePublishSuccess = successParam === 'published' && casePublished;
+		const caseUnpublishSuccess = successParam === 'unpublish' && !casePublished;
```
oscargay: These are never used now, I think that's why linting is failing?

apps/manage/src/app/views/cases/view/view.njk
```diff
@@ -97,6 +97,49 @@
...
+                        value: {
+                            text: lastModifiedDate if lastModifiedDate else "-"
+                        }
+                    },
+                    {
+                        key: {
+                            text: "Last modified by"
+                        },
+                        value: {
+                            text: lastModifiedBy if lastModifiedBy else "-"
```
oscargay: If you do this in the controller you don't need the `if-else` here anymore?

apps/manage/src/app/views/cases/case-history/view-model.ts
pinssbrett: I don't think `resolveTemplate` does handle unknown actions safely - it will throw an error when it tries to read `undefined.replace()`. Because its input type is `AuditAction`, it shouldn't need to, so long as that action is validated somewhere upstream. Otherwise forcing unsafe values into a type we've defined sort of defeats the purpose of typing it. Happy to discuss this more though!

pinssbrett (commented):
I added further discussion about the AuditAction cast

pinssbrett (approved):
Nice, looks good!

PR #985, opened 2026-05-26
Feat/peas 6 audit create case event

paymonsattar:
## Describe your changes

Audits the create a case event - this is an example implementation of auditing.

Screenshot 2026-05-26 at 17 01 54
Screenshot 2026-05-26 at 17 02 11

### DIFF
--- a/apps/manage/src/app/views/cases/case-history/view.njk
+++ b/apps/manage/src/app/views/cases/case-history/view.njk
@@ -24,6 +24,7 @@
                     <p class="govuk-body" data-cy="no-history-message">There is no case history to display</p>
                 {% else %}
                     <p class="govuk-body govuk-clearfix">
+<<<<<<< HEAD
                         <span class="govuk-body" data-cy="pagination-results-text">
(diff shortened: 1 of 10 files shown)


apps/manage/src/app/views/cases/create-a-case/save.js
```diff
@@ -96,10 +97,19 @@ export function buildSaveController(service) {
 			}
 		});
 
-		if (!reference || !lbcReference) {
+		if (!reference || !lbcReference || !id) {
```
oscargay: We should probably handle this independently because the error message is misleading. I'm guessing this check is because TS complains that technically `id` could be undefined due to the `let`?

oscargay (commented):
Looks good, 1 minor comment about a potentially misleading error handler message !

apps/manage/src/app/views/cases/case-history/view.njk
```diff
@@ -65,7 +65,7 @@
 
                         {% set tableRows = (tableRows.push([
                             {
-                                html: row.dateTime,
+                                html: row.dateTimeFormatted,
```
paymonsattar, line 68: forgot to update this field in the last PR

AlexMoulin (approved):
LGTM

PR #989, opened 2026-05-29
feat: first pass at adding azure blob to crown

oscargay:
## Describe your changes
- Adds the skeleton of Azure blob into the Crown repo.
- Not wired up to anything, and is set to `disabled` in the terraform files for all the servers so it shouldn't get instantiated.

[PEAS-41](https://pins-ds.atlassian.net/browse/PEAS-41)

[PEAS-41]: https://pins-ds.atlassian.net/browse/PEAS-41?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/manage/src/app/config.js
+++ b/apps/manage/src/app/config.js
@@ -27,7 +27,7 @@ export function loadConfig() {
 	}
 	// load configuration from .env file into process.env
 	// prettier-ignore
-	try { loadEnvFile() } catch {/* ignore errors here */}
+	try { loadEnvFile() } catch { /* ignore errors here */ }
 
 	// get values from the environment
 	const {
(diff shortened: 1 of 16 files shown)


packages/lib/blob-store/blob-store-client.ts
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	private getFileUploadOptions(fileType: string): BlockBlobUploadStreamOptions {
+		return { blobHTTPHeaders: { blobContentType: fileType } };
+	}
+
+	async upload(buffer: Buffer, mimeType: string, blobName: string): Promise<BlobUploadCommonResponse> {
+		try {
+			const containerClient = this.blobServiceClient.getContainerClient(this.container);
+			await containerClient.createIfNotExists();
```
paymonsattar: Suggestion: You have this in multiple places - it would be better if we move this logic into it's own function. For example:

```
async getBlobContainerClient() {
		const containerClient = this.blobServiceClient.getContainerClient(this.container);
		await containerClient.createIfNotExists();

       return containerClient
}
```

apps/manage/src/app/config.js
```diff
@@ -143,6 +147,24 @@ export function loadConfig() {
...
+			BLOB_STORE_CONTAINER
+		};
+		for (const [k, v] of Object.entries(props)) {
+			if (v === undefined || v === '') {
+				throw new Error(k + ' must be a non-empty string');
+			}
+		}
+
+		if (NODE_ENV === 'production' && BLOB_STORE_CONNECTION_STRING) {
+			throw new Error(BLOB_STORE_CONNECTION_STRING + ' must only be used for local development');
```
paymonsattar: I'm not sure if this exists in MPESC too - but I would advise against outputting the connection string env var in the error message - could be a secret leak

oscargay: Good spot, think this just needs to be the name of the variable instead!

packages/lib/blob-store/blob-store-client.ts
oscargay: Have looked at this, there were a few other areas that I felt could be a little cleaner after you noted it so have done a bit of a refactor

paymonsattar (approved):

PR #990, opened 2026-05-29
fix(csrf): update dynamic-forms dependency and add CSRF token to cookie form

AlexMoulin:
## Describe your changes
Updates the cookie post to pass a csrf token and update dynamic forms for the fix to the check your answers page csrf error


oscargay (dismissed):
Looks good, only thoughts were checking nothing else breaking would be pulled across from `dynamic-forms` between the releases, but seems okay from what I can see!

apps/portal/src/app/views/static/cookies/view.njk
pinssbrett, line 1: I think this is still missing the field in the forms on `cookie-banner.njk`

pinssbrett (changes requested):
Need to handle cookie banner as well as static page

apps/portal/src/app/views/static/cookies/view.njk
AlexMoulin: Nice catch! I will do that now!

pinssbrett (approved):
Looks good!

PR #992, opened 2026-06-01
Feat/crown 1560 1561 1562 invalid state has agent

pinssbrett:
## Describe your changes
Add banner notification messages when case is in an invalid state. Covers three scenarios:
1. Case has agent but is missing agent organisation details
2. Case has agent but it missing agent contact details
3. Case does not have agent, and there is not at least one applicant contact for each applicant organisation

This also adds some standard payload contracts for database calls.

https://pins-ds.atlassian.net/browse/CROWN-1560
https://pins-ds.atlassian.net/browse/CROWN-1561
https://pins-ds.atlassian.net/browse/CROWN-1563
Screenshot 2026-06-01 093529
Screenshot 2026-06-01 093536
https://pins-ds.atlassian.net/browse/CROWN-1562
Screenshot 2026-06-01 090741
Screenshot 2026-06-01 090749
Screenshot 2026-06-01 090939
Screenshot 2026-06-01 093510

### DIFF
--- a/apps/manage/src/app/views/cases/view/controller.ts
+++ b/apps/manage/src/app/views/cases/view/controller.ts
@@ -19,6 +19,7 @@ import { APPLICATION_SUB_TYPE_ID, APPLICATION_TYPE_ID } from '@pins/crowndev-dat
 import { filteredStagesToRadioOptions } from './question-utils.js';
 import { getApplicantOrganisationOptions } from '../util/applicant-organisation-options.js';
 import { BannerBuilder } from '@pins/crowndev-lib/views/banner/banner-builder.ts';
+import { CROWN_DEVELOPMENT_VIEW_INCLUDE } from './payload-contracts.ts';
 import type { SharePointDrive } from '@pins/crowndev-sharepoint/src/sharepoint/drives/drives.js';
 import type { Response, Request, Handler, NextFunction } from 'express';
 import type { ErrorSummaryItem } from '@pins/crowndev-lib/util/types.ts';
@@ -205,31 +206,7 @@ export function buildGetJourneyMiddleware(service: ManageService, isQuestionView
 
 		const crownDevelopment = await db.crownDevelopment.findUnique({
 			where: { id },
-			include: {
-				ApplicantContact: { include: { Address: true } },
-				AgentContact: { include: { Address: true } },
-				Category: { include: { ParentCategory: true } },
-				Event: true,
-				Lpa: { include: { Address: true } },
-				SecondaryLpa: { include: { Address: true } },
-				SiteAddress: true,
-				ParentCrownDevelopment: { select: { reference: true } },
-				ChildrenCrownDevelopment: { select: { reference: true } },
-				Organisations: {
-					include: {
-						Organisation: {
-							include: {
-								Address: true,
-								OrganisationToContact: {
-									include: {
-										Contact: { include: { Address: true } }
-									}
-								}
-							}
-						}
-					}
-				}
-			}
+			include: CROWN_DEVELOPMENT_VIEW_INCLUDE
 		});
 		if (crownDevelopment === null) {
 			return notFoundHandler(req, res);
--- a/apps/manage/src/app/views/cases/view/controller.ts
+++ b/apps/manage/src/app/views/cases/view/controller.ts
@@ -19,7 +19,8 @@ import { APPLICATION_SUB_TYPE_ID, APPLICATION_TYPE_ID } from '@pins/crowndev-dat
 import { filteredStagesToRadioOptions } from './question-utils.js';
 import { getApplicantOrganisationOptions } from '../util/applicant-organisation-options.js';
 import { BannerBuilder } from '@pins/crowndev-lib/views/banner/banner-builder.ts';
-import { CROWN_DEVELOPMENT_VIEW_INCLUDE } from './payload-contracts.ts';
+import { escapeHtml } from '@pins/crowndev-lib/util/string.ts';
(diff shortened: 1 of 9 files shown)


apps/manage/src/app/views/cases/view/update-case.js
```diff
@@ -45,6 +46,21 @@ export function buildUpdateCase(service, clearAnswer = false) {
 
 		await customUpdateCaseActions(service, id, toSave, fullViewModel);
 
+		if (toSave.hasAgent === false && fullViewModel.hasAgent === 'yes') {
```
AlexMoulin, line 49: IIRC fullViewModel gets mutated which is why when we are looking at the original answers we use originalAnswers. I want to fix this but last time I look there was quite a bit of entanglement so didn't get it done due to other priorities.

packages/lib/views/banner/banner-builder.ts
```diff
@@ -42,7 +42,7 @@ const HTML_ESCAPES: Record<string, string> = {
 /**
  * Escape a string so it can be safely rendered as HTML text.
  */
-function escapeHtml(value: string): string {
+export function escapeHtml(value: string): string {
```
AlexMoulin: Now that this is exported, it makes me think it could be a really useful util to have across the app.

apps/manage/src/app/views/cases/view/controller.ts
```diff
@@ -31,6 +32,65 @@ function getJourneyAnswers(res: Response): CrownDevelopmentViewModel | undefined
...
+		missingInfo.push('Agent contact');
+	}
+
+	if (missingInfo.length === 0) {
+		return null;
+	}
+
+	return `<p class="govuk-body">${applicantContactText}:</p>
+			<ul class="govuk-list govuk-list--bullet">
+			<li>${missingInfo.join('</li><li>')}</li>
```
AlexMoulin, line 91: I have never done it this way, very cool!

AlexMoulin (changes requested):
Just need to move the escapeHtml

apps/manage/src/app/views/cases/view/controller.ts
```diff
@@ -70,6 +131,22 @@ async function getBannerMessages(id: string, res: Response, req: Request, db: Ma
 		return bannerBuilder.build();
 	}
 
+	const crownDevelopmentView = res.locals.originalAnswers;
+
+	const agentStatusUpdated = readSessionData(req, id, 'agentStatusUpdated', false);
```
paymonsattar, line 137: You might be introducing a race condition here, could be worth clearing the sessions after you use them in the if statement on line 141... or even in a try...finally catch. 

What do you think?

apps/manage/src/app/views/cases/view/controller.ts
```diff
@@ -31,6 +32,65 @@ function getJourneyAnswers(res: Response): CrownDevelopmentViewModel | undefined
...
+	const applicantContactText = options.relatedFieldUpdated
+		? 'You need to add'
+		: 'This application is missing information';
+
+	const missingInfo: string[] = [];
+
+	/**
+	 * If no agent, each applicant organisation must have at least one contact.
+	 */
+	if (orgsWithoutContacts && orgsWithoutContacts.length > 0 && crownDevelopmentView.hasAgent === 'no') {
```
paymonsattar, line 65: Because `hasAgent` is a string, if the value is neither `'yes'` nor `'no'` then both the if statements are skipped and it could potentially hide an invalid state. If this is by purpose could we add a comment somewhere please to explain it and return null:

```
// If hasAgent is not yet set, we can't determine validity — skip the check
if (!crownDevelopmentView.hasAgent) return null;
```

Or (and preferably) make the intent explicit:
`const hasAgent = crownDevelopmentView.hasAgent ?? 'no';`

apps/manage/src/app/views/cases/view/update-case.js
paymonsattar, line 49: Suggestion: There is an asymmetry concern for me here, although this works - it's a little ugly to look at. 

It would look a lot cleaner like this:

```
const boolToYesNo = (val) => val ? 'yes' : 'no';

if (boolToYesNo(toSave.hasAgent) !== fullViewModel.hasAgent) {
    addSessionData(req, id, { agentStatusUpdated: true });
}
```

paymonsattar (changes requested):
I've left a few comments, let me know what you think

apps/manage/src/app/views/cases/view/controller.ts
pinssbrett: not sure if it would be a race condition, but agree it will be a bit clearer this way - i have updated

apps/manage/src/app/views/cases/view/controller.ts
pinssbrett: the type of `hasAgent` is narrower than `string`: it's never optional and can only be `"yes" | "no"`. so I'm satisfied that no extra check is required here

apps/manage/src/app/views/cases/view/update-case.js
pinssbrett: yeah it looks awful. however i feel like intentionally depicting the schematic differences between the view model and the save model is the clearest way of representing it - converting it just for comparison might hide the actual structure. the ugly truth!

apps/manage/src/app/views/cases/view/controller.ts
paymonsattar: Sounds good - if it's always either `yes` or `no` then I'm happy with this approach

paymonsattar (approved):
LGTM

AlexMoulin (approved):
LGTM

PR #995, opened 2026-06-01
feat(manage): combine banner messages for representation management

pinssbrett:
## Describe your changes
Part of banner combination: combine all banners on manage reps endpoint. Here the messages are changes in representation status, and if the a representation contains distressing content but the case is not marked as doing so.

https://pins-ds.atlassian.net/browse/CROWN-1558
Screenshot 2026-06-01 163643


paymonsattar (approved):
LGTM

PR #998, opened 2026-06-02
feat(update-case): convert update-case.js to TypeScript

pinssbrett:
## Describe your changes
Convert `update-case` to typescript.

I've made some updates in how numbers are handled in `CrownDevelopmentSaveModel`. Essentially if a `number` question is used, these are `number` in both schemas. But some of our number fields use `multi-field-input` (e.g. site coordinates), which retains them as `string` in the save model. Therefore I've:
- separated out these fields so only they are `number` in the view model and `string` in the save model
- removed redundant number conversion for application fee field that is already a number
- added [a ticket to the backlog](https://pins-ds.atlassian.net/browse/CROWN-1620) for improving the multi-field input so these can be treated as the number fields they are.

Clearing data now also has an extra check to make sure that required fields are not clearable (a double-check, since these should also be required questions).

N/A


paymonsattar (dismissed):
LGTM

AlexMoulin (approved):
LGTM

PR #999, opened 2026-06-02
feat(manage): delete agent org and contacts when hasAgent changes to no

pinssbrett:
## Describe your changes
Handle deletion of organisations, contacts and addresses when `hasAgent` is changed from yes to no:

* Update `buildCaseUpdateWritePlan` in `linked-case-updates.ts` to detect when `hasAgent` changes from "yes" to "no" and queue deletions for all related agent organisations, contacts, addresses, and links across all relevant cases.
* Extend the `CaseUpdateWritePlan` type to include new fields for bulk deletion of agent links, contacts, organisations, and addresses.
* Extend `executeCaseUpdateWritePlan` to process and execute the new bulk deletion instructions for agent data.
* Add tests for the above

https://pins-ds.atlassian.net/browse/CROWN-1564
Before deletion: agent details containing "1325" exist in DB tables:
Screenshot 2026-06-02 132610
Screenshot 2026-06-02 132650
Screenshot 2026-06-02 132658
Screenshot 2026-06-02 132705
After deletion: these DB entries have all been removed:
Screenshot 2026-06-02 132727
Screenshot 2026-06-02 132732
Screenshot 2026-06-02 132738
And when setting `hasAgent` back to yes, the old data does not persist:
Screenshot 2026-06-02 132758

### DIFF
--- a/apps/manage/src/app/views/cases/view/linked-case-updates.ts
+++ b/apps/manage/src/app/views/cases/view/linked-case-updates.ts
@@ -27,6 +27,9 @@ type CaseUpdateWritePlan = {
 		newOrganisationContacts: Array<{ organisationId: string; data: Prisma.ContactCreateInput }>;
 		organisationToContactDeletes: Array<{ id: string }>;
 		organisationToContactCreates: Array<{ organisationId: string; contactId: string }>;
+		agentLinkDeleteMany?: Array<{ caseIds: string[]; roleId: string; organisationId: string }>;
+		agentContactDeleteMany?: Array<{ contactIds: string[] }>;
+		agentOrganisationDeleteMany?: Array<{ organisationIds: string[] }>;
 	};
 };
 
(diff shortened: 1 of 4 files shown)


apps/manage/src/app/views/cases/view/linked-case-updates.ts
```diff
@@ -413,6 +417,55 @@ export function buildCaseUpdateWritePlan({
 		}
 	}
 
+	// When hasAgent is set from yes to no, remove all agent joins, case links, contacts, and organisations.
+	if (toSave.hasAgent === false && dbViewModel.hasAgent === 'yes') {
```
paymonsattar: Could you use `BOOLEAN_OPTIONS` instead of writing out the raw 'yes' string - You've used the object in [‎apps/manage/src/app/views/cases/view/update-case.js](https://github.com/Planning-Inspectorate/crown-developments/pull/999/changes#diff-550b9c354792f18e6f8531263642e52f6484b912048e9c3fc9a2e4e0e5fd60a0) so would be good to keep it consistant

paymonsattar (changes requested):
Small comment - but other than that all good

apps/manage/src/app/views/cases/view/linked-case-updates.ts
AlexMoulin: Agreed, best practice would be to use the constants.

AlexMoulin (changes requested):
Minor nitpick to address but other than that looks good @pinssbrett

paymonsattar (approved):

AlexMoulin (approved):

PR #1000, opened 2026-06-03
feat(manage): add anchor link to audit section

paymonsattar:
## Describe your changes
Adds an anchor link to the top of the case details page to navigate to the audit section. Have hardcoded this in rather than adding at the controller level for two reasons:

1) The section is currently hardcoded and changing it to be loaded from the controller would be a bigger change and more fiddly.
2) We want the audit section to always appear last.

Screenshot 2026-06-03 at 13 42 57
Screenshot 2026-06-03 at 13 43 10
number and link

## Issue ticket 
https://pins-ds.atlassian.net/browse/CROWN-1625

### DIFF
--- a/apps/manage/src/app/views/cases/view/view.njk
+++ b/apps/manage/src/app/views/cases/view/view.njk
@@ -44,6 +44,9 @@
                         </li>
                     {% endif %}
                 {% endfor %}
+                <li>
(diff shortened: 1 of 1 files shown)


pinssbrett (approved):
Makes sense, since defined separately from question sections

PR #1013, opened 2026-06-08
fix(session): add isUnsafeObjectKey check to prevent prototype pollution

AlexMoulin:
## Describe your changes
This pull request introduces enhanced session security by preventing the use of unsafe object keys (such as `__proto__`, `prototype`, and `constructor`) when interacting with session data across several controllers and utility files. The changes add explicit checks using a new `isUnsafeObjectKey` utility, throwing errors or early-returning when unsafe keys are detected. This mitigates prototype pollution vulnerabilities and improves the overall robustness of session handling.

**Session Security Improvements:**

* Added and used the `isUnsafeObjectKey` function from `@pins/crowndev-lib/util/session.ts` to check for unsafe object keys before reading, writing, or deleting session data in multiple controllers and utility files. [[1]](diffhunk://#diff-3c24285a24f3e85c90121a377e288258a25830fdd0ebb214a51b99dd9090e69cL1-R6) [[2]](diffhunk://#diff-f8df097ea116aea6bde7351fa0715c816366779f631d652927f97441d2551bd3L15-R15) [[3]](diffhunk://#diff-2190c3c5b39c7bf3c7d27b5e16b95bdf8dec5629942af9773236412779ca86a5L4-R9) [[4]](diffhunk://#diff-16d50753971935f089363d927bf571e8343a04faaea69b550071356a86950318L9-R13)
* Updated session handling functions such as `getApplicationUpdateSessionData`, `clearAppUpdatesFromSession`, `readCaseUpdatedSession`, `clearCaseUpdatedSession`, `updateRepReviewSession`, and `safeDeleteUploadedFilesSession` to throw errors or return early if an unsafe key is detected. [[1]](diffhunk://#diff-3c24285a24f3e85c90121a377e288258a25830fdd0ebb214a51b99dd9090e69cR35-R37) [[2]](diffhunk://#diff-3c24285a24f3e85c90121a377e288258a25830fdd0ebb214a51b99dd9090e69cL47-R57) [[3]](diffhunk://#diff-f8df097ea116aea6bde7351fa0715c816366779f631d652927f97441d2551bd3L250-R253) [[4]](diffhunk://#diff-f8df097ea116aea6bde7351fa0715c816366779f631d652927f97441d2551bd3R265-R269) [[5]](diffhunk://#diff-2190c3c5b39c7bf3c7d27b5e16b95bdf8dec5629942af9773236412779ca86a5R788-R791) [[6]](diffhunk://#diff-2190c3c5b39c7bf3c7d27b5e16b95bdf8dec5629942af9773236412779ca86a5R1053-R1055)
* Updated filtering logic in `updateDocumentStatusSession` to skip unsafe keys, further protecting session data from prototype pollution.

**Code Quality and Consistency:**

* Refactored loops and assignments to use safer iteration patterns (e.g., `Object.entries` instead of `for...in`) when working with session data.
* Updated the implementation of `addCaseUpdatedSession` to use the safer `addSessionData` helper instead of direct session mutation.

**Utility Enhancements:**

* Exported the `isUnsafeObjectKey` function in `@pins/crowndev-lib/util/session.ts` and documented its purpose.
CodeQL #58


paymonsattar (dismissed):

apps/manage/src/app/views/cases/view/application-updates/utils.js
```diff
@@ -27,6 +32,9 @@ export function clearAppUpdateStatusSession(req, id) {
 }
 
 export function getApplicationUpdateSessionData(req, applicationUpdateId) {
+	if (typeof applicationUpdateId === 'string' && isUnsafeObjectKey(applicationUpdateId)) {
```
oscargay: Not sure if this is going to be too fiddly to do now, but it could be worth having these sorts of checks in a middleware somewhere? I see there is a `validateParams` function above, could these safety checks be included when we check for the existence of them as well? Could make it easier to extend, reduce the repetition, and also avoid running code that is ultimately going to hit an error further down the line?

pinssbrett: I guess this is a good candidate for TS conversion soon, since this should always be a string

pinssbrett (dismissed):
Looks good!

oscargay (commented):
Looks good, just thinking about whether there's a way to make these sorts of checks more generic / reusable alongside other validation middleware we do on parameters? Let me know what you think

apps/manage/src/app/views/cases/view/application-updates/utils.js
AlexMoulin: Fair point! I will reassess my approach.

AlexMoulin: I can do that :)

oscargay (approved):
lgtm

PR #1017, opened 2026-06-08
feat(manage): combine banner messages for application updates

pinssbrett:
## Describe your changes
Use banner builder to standardise banners on manage application updates page.

https://pins-ds.atlassian.net/browse/CROWN-1558
image

### DIFF
--- a/apps/manage/src/app/views/cases/view/application-updates/controller.js
+++ b/apps/manage/src/app/views/cases/view/application-updates/controller.js
@@ -10,6 +10,28 @@ import {
 } from './utils.js';
 import { APPLICATION_UPDATE_STATUS_ID } from '@pins/crowndev-database/src/seed/data-static.ts';
 import { BOOLEAN_OPTIONS } from '@planning-inspectorate/dynamic-forms/src/components/boolean/question.js';
+import { BannerBuilder } from '@pins/crowndev-lib/views/banner/banner-builder.ts';
+
+/**
+ * @typedef {import('@pins/crowndev-lib/views/banner/banner-builder').BannerMessage} BannerMessage
+ */
+
+/**
+ * Get all banner messages to display.
+ *
+ * @param {Object} options
+ * @param {string|false} options.applicationUpdateStatus
+ * @return {BannerMessage|null}
+ */
+function getBannerMessages({ applicationUpdateStatus }) {
+	const bannerBuilder = new BannerBuilder();
+
+	if (applicationUpdateStatus) {
+		bannerBuilder.addSuccessText(applicationUpdateStatus);
+	}
+
+	return bannerBuilder.build();
+}
 
 export function buildApplicationUpdates({ db }) {
 	return async (req, res) => {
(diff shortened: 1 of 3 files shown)


apps/manage/src/app/views/cases/view/application-updates/view.njk
```diff
@@ -8,11 +8,12 @@
...
-        {{ govukNotificationBanner({
-            text: applicationUpdateStatus,
-            type: "success"
-        }) }}
+    {% if banner %}
+        <div class="govuk-grid-row">
+            <div class="govuk-grid-column-full">
+                {{ govukNotificationBanner(banner) }}
+            </div>
+        </div>
```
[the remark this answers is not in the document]
pinssbrett: Out of scope

apps/manage/src/app/views/cases/view/application-updates/controller.js
```diff
@@ -10,6 +10,31 @@ import {
...
+ *
+ * @param {string} id
+ * @param {import('express').Request} req
+ * @return {BannerMessage|null}
+ */
+function getBannerMessages(id, req) {
+	const bannerBuilder = new BannerBuilder();
+	const applicationUpdateStatus = readAppUpdateStatus(req, id);
+	clearAppUpdateStatusSession(req, id);
+	clearAppUpdatesFromSession(req);
```
oscargay: I think having this logic inside of this function sort of gives it 2 purposes, and also makes it a little harder to test things like this because it will be working with requests and sessions?

To me it would make it a little cleaner to have this outside of the function call, and then pass it in as a parameter? But it's a relatively small point, what do you think?

```javascript
		const applicationUpdateStatus = readAppUpdateStatus(req, id);
		clearAppUpdateStatusSession(req, id);
		clearAppUpdatesFromSession(req);
		const banner = getBannerMessages(id, req, applicationUpdateStatus);

```

oscargay (commented):
Cool, looks good, just 1 comment, let me know your thoughts!

apps/manage/src/app/views/cases/view/application-updates/controller.js
pinssbrett: Good point, I agree!

apps/manage/src/app/views/cases/view/application-updates/controller.js
```diff
@@ -10,6 +10,28 @@ import {
...
+
+/**
+ * @typedef {import('@pins/crowndev-lib/views/banner/banner-builder').BannerMessage} BannerMessage
+ */
+
+/**
+ * Get all banner messages to display.
+ *
+ * @param {Object} options
+ * @param {string|null} options.applicationUpdateStatus
```
oscargay: Very small one, but based on `readAppUpdateStatus` I think this is `string|false` rather than `null` - could be wrong on this one, just basing this on `false` being the default fallback when reading from session

oscargay (approved):
LGTM!

PR #1024, opened 2026-06-09
Feat/convert custom components index to typescript

pinssbrett:
## Describe your changes
Convert `custom-components` index file to typescript:

- Add new `CrownQuestionProps` type that encompasses all custom and dynamic-forms question properties
- Redefine custom question properties in the file so that they are accessible to typescript - when we convert the questions to TS, these can be within the question files themselves
- Change a `null` return statement in `RepresentationAttachments.getAction()` to `undefined` to match method return type
- Update `dynamic-forms` to include some recent type fixes

N/A


RobBroadhead30 (approved):
Looks good to me :)

PR #1025, opened 2026-06-09
fix(manage): add regex validation for agent first and last name fields

RobBroadhead30:
## Describe your changes

This pull request enhances validation for contact name fields in the case creation flow by enforcing stricter input rules and adding comprehensive tests to ensure correct behavior. The changes ensure that first and last names only allow letters, spaces, hyphens, and apostrophes, and that this is thoroughly tested for different contact types.

**Validation improvements:**

* Updated the `multiContactQuestions` function in `question-utils.js` to add a regex validator for both first and last name fields, restricting input to only letters, spaces, hyphens, and apostrophes, and providing a clear error message if the input does not match.

**Testing enhancements:**

* Added new tests in `question-utils.test.js` to verify that the regex validation for first and last names is correctly applied, rejects numbers, and is enforced for both agent and applicant contacts.

https://pins-ds.atlassian.net/browse/CROWN-1607
image

### DIFF
--- a/apps/manage/src/app/views/cases/create-a-case/question-utils.test.js
+++ b/apps/manage/src/app/views/cases/create-a-case/question-utils.test.js
@@ -185,4 +185,81 @@ describe('multiContactQuestions', () => {
 
 		assert.strictEqual(organisationFieldRule, undefined);
 	});
+
+	describe('name field validation', () => {
+		const questions = multiContactQuestions({
+			prefix: 'agent',
+			title: 'agent',
+			organisationOptions: null
+		});
+
+		const question = questions.agentContactDetails;
+		const [multiFieldValidator] = question.validators;
+
+		it('should validate first name with regex allowing only letters, spaces, hyphens and apostrophes', () => {
+			const firstNameFieldRule = multiFieldValidator.fields.find(
+				(fieldRule) => fieldRule.fieldName === 'agentFirstName'
+			);
+
+			assert.ok(firstNameFieldRule);
+			assert.ok(Array.isArray(firstNameFieldRule.validators));
+
+			const stringValidator = firstNameFieldRule.validators.find(
+				(validator) => validator.constructor.name === 'StringValidator'
+			);
+
+			assert.ok(stringValidator);
+			assert.ok(stringValidator.regex);
+			assert.strictEqual(
+				stringValidator.regex.regexMessage,
+				'First name must only include letters, spaces, hyphens and apostrophes'
+			);
+
+			const regex = new RegExp(stringValidator.regex.regex);
+
+			assert.ok(regex.test('Jane'));
+			assert.ok(regex.test('Mary Jane'));
+			assert.ok(regex.test("O'Brien"));
+			assert.ok(regex.test('Anne-Marie'));
+			assert.ok(regex.test("Mary-Jane O'Connor"));
+
+			assert.ok(!regex.test('Jane2'));
+			assert.ok(!regex.test('Jane!'));
+			assert.ok(!regex.test('Mary@Jane'));
+			assert.ok(!regex.test('Mary.Jane'));
+			assert.ok(!regex.test(''));
+		});
+
+		it('should validate last name with regex allowing only letters, spaces, hyphens and apostrophes', () => {
+			const lastNameFieldRule = multiFieldValidator.fields.find((fieldRule) => fieldRule.fieldName === 'agentLastName');
+
+			assert.ok(lastNameFieldRule);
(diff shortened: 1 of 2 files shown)


apps/manage/src/app/views/cases/create-a-case/question-utils.test.js
```diff
@@ -185,4 +185,94 @@ describe('multiContactQuestions', () => {
...
+		assert.ok(firstNameStringValidator);
+		assert.strictEqual(firstNameStringValidator.regex.regex, "^[A-Za-z ''-]+$");
+
+		// Check last name
+		const lastNameStringValidator = lastNameFieldRule.validators.find(
+			(validator) => validator.constructor.name === 'StringValidator'
+		);
+		assert.ok(lastNameStringValidator);
+		assert.strictEqual(lastNameStringValidator.regex.regex, "^[A-Za-z ''-]+$");
+	});
```
pinssbrett: This test is redundant - because the function generates the same fields for any prefix, your `agent` tests above are sufficient

apps/manage/src/app/views/cases/create-a-case/question-utils.test.js
```diff
@@ -185,4 +185,94 @@ describe('multiContactQuestions', () => {
...
+		assert.ok(Array.isArray(lastNameFieldRule.validators));
+
+		// Find the StringValidator
+		const stringValidator = lastNameFieldRule.validators.find(
+			(validator) => validator.constructor.name === 'StringValidator'
+		);
+
+		assert.ok(stringValidator);
+		assert.ok(stringValidator.regex);
+		assert.strictEqual(stringValidator.regex.regex, "^[A-Za-z ''-]+$");
```
pinssbrett: Although this checks presence of the regex itself, it doesn't actually test the behaviour. Should we add some tests for example names in order to test this?

pinssbrett (changes requested):
A few tweaks to tests required

apps/manage/src/app/views/cases/create-a-case/question-utils.test.js
```diff
@@ -185,4 +185,116 @@ describe('multiContactQuestions', () => {
...
+			assert.ok(!regex.test('Mary.Jane'));
+			assert.ok(!regex.test('John_Smith'));
+			assert.ok(!regex.test('Sarah!'));
+			assert.ok(!regex.test('John,Smith'));
+		});
+
+		it('should reject empty strings', () => {
+			assert.ok(!regex.test(''));
+		});
+	});
```
pinssbrett: This is not testing anything from the codebase, just the regex set within the test.
In order to test the regex in the codebase, instead modify the existing tests to use what is returned by the validator. For example:
```
    assert.ok(stringValidator);
    assert.ok(stringValidator.regex);

    const regex = new RegExp(stringValidator.regex.regex);

    // valid names
    assert.ok(regex.test('Jane'));
    assert.ok(regex.test("O'Brien"));
    assert.ok(regex.test('Smith-Jones'));

    // invalid names
    assert.ok(!regex.test('Jane2'), 'should reject numbers');
    assert.ok(!regex.test('Jane!'), 'should reject special characters');
```

pinssbrett (changes requested):
Make sure to test the regex in the code itself

apps/manage/src/app/views/cases/create-a-case/question-utils.test.js
```diff
@@ -185,4 +185,83 @@ describe('multiContactQuestions', () => {
...
+			assert.ok(firstNameFieldRule);
+			assert.ok(Array.isArray(firstNameFieldRule.validators));
+
+			const stringValidator = firstNameFieldRule.validators.find(
+				(validator) => validator.constructor.name === 'StringValidator'
+			);
+
+			assert.ok(stringValidator);
+			assert.ok(stringValidator.regex);
+			assert.strictEqual(stringValidator.regex.regex, "^[A-Za-z ''-]+$");
```
pinssbrett: Because we are testing the behaviour of the regex we specifically don't want to test the regex string itself - instead all functionality gets tested by the behaviour tests and is not strictly tied to the exact implementation. This means that any future changes to the validation can be understood as, or extended from, how the regex behaves and not the specific implementation (+ same for test below)
```suggestion
			assert.ok(stringValidator.regex);
```

pinssbrett (changes requested):
Did not quite explain myself fully on the previous review - one last change! Thanks

pinssbrett (approved):
Looks good!

PR #1029, opened 2026-06-11
feat(portal): add service link to header

RobBroadhead30:
## Describe your changes
This pull request updates the header navigation across the application to ensure the service name in the header links to the correct URL. The main changes involve passing a new `headerUrl` configuration value from the backend to the frontend template and using it in the header component.

Header navigation improvements:

* Added a `headerUrl` property to the locals configuration in `config-middleware.js`, setting it to `/applications` so the header title links to the applications page.
* Updated the `pins-header.njk` template to use the new `headerUrl` property for the service name link in the header navigation.
https://pins-ds.atlassian.net/browse/CROWN-1651
image

### DIFF
--- a/apps/portal/src/util/config-middleware.js
+++ b/apps/portal/src/util/config-middleware.js
@@ -29,6 +29,7 @@ export function addLocalsConfiguration({
 			appName,
 			cspNonce: res.locals.cspNonce,
 			headerTitle: 'Find a Crown development application',
+			headerUrl: '/',
 			footerLinks: [
 				{
 					text: 'Terms and conditions',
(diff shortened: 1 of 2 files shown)


apps/portal/src/util/config-middleware.js
```diff
@@ -29,6 +29,7 @@ export function addLocalsConfiguration({
 			appName,
 			cspNonce: res.locals.cspNonce,
 			headerTitle: 'Find a Crown development application',
+			headerUrl: '/applications',
```
pinssbrett: Even though it redirects to `applications`, the [instruction I got](https://pins-ds.atlassian.net/browse/CROWN-1613?focusedCommentId=121291) was that this should be the root path. I guess for future-proofing

```suggestion
			headerUrl: '/',
```

pinssbrett (changes requested):
minor tweak to tiny pr!

pinssbrett (approved):
👍

PR #1034, opened 2026-06-15
chore: converts gov-notify to typescript

oscargay:
## Describe your changes
- Converts GovNotify module to TypeScript

CROWN-1654


packages/lib/govnotify/gov-notify-client.ts
```diff
@@ -0,0 +1,234 @@
...
+				this.logger.error({ message: error.response.data.errors }, 'received from Notify API');
+			}
+
+			const errorMessage = error instanceof Error ? error.message : 'Unknown error';
+
+			throw new Error(`email failed to dispatch: ${errorMessage}`, { cause: error });
+		}
+	}
+
+	async getNotificationById(notificationId: string): Promise<unknown> {
```
pinssbrett: Is it possible to get the full return type from the package? Or add a TODO if it's not worth handling right now
```suggestion
	async getNotificationById(notificationId: string): Promise<AxiosResponse<NotificationResponse>> {
```

pinssbrett (dismissed):
Looks good! I only have a very minor type comment, which could be addressed in the future

packages/lib/govnotify/gov-notify-client.ts
oscargay: Ah good point, will have a look now, seeing as we're taking the time to TS it now might as well do the best job we can. Will investigate and see if we can grab the full response!

pinssbrett (approved):

PR #1039, opened 2026-06-17
fix(manage): don't allow auth redirects to other domains

beejjacobs:

https://pins-ds.atlassian.net/browse/RDSD-258


pinssbrett (approved):
Looks good to me

PR #1042, opened 2026-06-18
refactor(macros): abstract nunjuck case list table macros to shared lib

andrewlilley-solirius:
## Describe your changes
Refactored case list table formatting macros (Portal/S62A Portal: _crownDevelopmentsTable_, Manage: _casesTable_), and associated functional macro (_insertWbr_).

New macro (_genericCaseListTable_) in _/packages/lib/views/tables/tables.njk_ :

- Takes inputs (_developments, headers, columns, isSortable_)
- Displays as formatted _govukTable_

~Additional edits to BaseConfig variable names to standardise from cacheControl to staticCacheControl or dynamicCacheControl~

~BLOCKED - Awaiting merge of shared view model on ticket PEAS-9~

PEAS-13 Abstracted Logic Ticket
[https://pins-ds.atlassian.net/browse/PEAS-13](https://pins-ds.atlassian.net/browse/PEAS-13)

### DIFF
--- a/apps/manage/src/app/views/cases/list/controller.js
+++ b/apps/manage/src/app/views/cases/list/controller.js
@@ -1,4 +1,5 @@
 import { createWhereClause, splitStringQueries } from '@pins/crowndev-lib/util/search-queries.js';
+import { insertWbr, formatStatusTag } from '@pins/crowndev-lib/util/string.ts';
 
 /**
  * @param {import('#service').ManageService} service
@@ -51,9 +52,15 @@ export function crownDevelopmentToViewModel(crownDevelopment) {
 		id: crownDevelopment.id,
 		reference: crownDevelopment.reference,
 		lpaName: crownDevelopment.Lpa?.name,
-		status: crownDevelopment.Status?.displayName,
+		status: formatStatusTag(crownDevelopment.Status?.displayName),
 		type: crownDevelopment.Type?.displayName,
-		location: ''
+		location: '',
+		referenceLink:
+			'<a class="govuk-link" href="/cases/' +
+			crownDevelopment.id +
+			'">' +
+			insertWbr(crownDevelopment.reference) +
+			'</a>'
 	};
 	if (crownDevelopment.SiteAddress) {
 		fields.location = addressToViewModel(crownDevelopment.SiteAddress);
--- a/apps/manage/src/app/views/cases/list/view.njk
+++ b/apps/manage/src/app/views/cases/list/view.njk
@@ -1,9 +1,9 @@
 {% extends "views/layouts/main.njk" %}
 
 {% from "govuk/components/button/macro.njk" import govukButton %}
-{% from "govuk/components/table/macro.njk" import govukTable %}
 {% from "govuk/components/tag/macro.njk" import govukTag %}
 {% from "search-bar/search-bar.njk" import searchBar %}
+{% from "tables/tables.njk" import baseCaseListTable, baseManageHeaders, baseManageColumns %}
 
 {% block pageTitle %}
     Case list - All Cases - Manage a Crown Development Application
@@ -50,75 +50,8 @@
         <p class="govuk-body">No results were found.</p>
         <p class="govuk-body"><a href="/cases" class="govuk-link govuk-link--no-visited-state">Clear search</a> to view all cases.</p>
     {% else %}
-        {{ casesTable(crownDevelopments) }}
+        {{ baseCaseListTable(crownDevelopments,baseManageHeaders,baseManageColumns,true) }}
     {% endif %}
 {% endblock %}
 
-{% macro casesTable(developments) %}
-    {% set developments = [] %}
 
-    {% for development in crownDevelopments %}
-        {% set developments = (developments.push([
-            {
-                html: "<a class='govuk-link' href='/cases/" +  development.id +"'>" + development.reference + "</a>",
-                attributes: {
-                    "data-sort-value": development.reference
-                }
-            },
-            {
-                text: development.location,
-                classes: 'pins-white-space-pre-wrap'
-            },
-            {
-                text: development.lpaName
-            },
-            {
-                text: development.type
-            },
-            {
-                html: govukTag({text: development.status})
-            }
-        ]), developments) %}
-    {% endfor %}
-
-    {% set headers = [
-        {
-            text: 'Reference',
-            attributes: {
-            "aria-sort": "descending"
-            }
-        },
-        {
-            text: 'Site location',
-            attributes: {
-            "aria-sort": "none"
-            }
-        },
-        {
-            text: 'Local planning authority (LPA)',
-            attributes: {
-            "aria-sort": "none"
-            }
-        },
-        {
-          text: 'Application type',
-            attributes: {
-            "aria-sort": "none"
-            }
-        },
-        {
-            text: 'Status',
-            attributes: {
-            "aria-sort": "none"
-            }
-        }
-    ] %}
-
-    {{ govukTable({
-        attributes: {
-            "data-module": "moj-sortable-table"
-        },
-        head: headers,
-        rows: developments
-    }) }}
-{% endmacro %}
--- a/apps/manage/src/app/views/s62a/cases/list/view-model.ts
+++ b/apps/manage/src/app/views/s62a/cases/list/view-model.ts
@@ -1,6 +1,7 @@
 import type { Prisma } from '@pins/crowndev-database/src/client/client.ts';
 import { addressToViewModel } from '@pins/crowndev-lib/util/address.ts';
 import { ORGANISATION_ROLES_ID } from '@pins/crowndev-database/src/seed/data-static.ts';
+import { insertWbr, formatStatusTag } from '@pins/crowndev-lib/util/string.ts';
 
 export interface S62ACaseView {
 	id: string;
@@ -38,14 +39,15 @@ export function s62aToViewModel(s62aCases: S62ACasePayload) {
 		id: s62aCases.id,
 		reference: s62aCases.reference,
 		lpaName: s62aCases.Lpa?.name,
-		status: s62aCases.S62aStatus?.displayName ?? undefined,
+		status: formatStatusTag(s62aCases.S62aStatus?.displayName),
 		type: s62aCases.Type?.displayName ?? undefined,
 		applicantOrganisations:
 			s62aCases.S62aToApplicants?.filter(
 				(relation) =>
 					relation.roleId === ORGANISATION_ROLES_ID.APPLICANT && typeof relation.Organisation?.name === 'string'
 			).map((item) => item.Organisation!.name) ?? [],
-		location: ''
+		location: '',
+		referenceLink: '<a class="govuk-link" href="/cases/' + s62aCases.id + '">' + insertWbr(s62aCases.reference) + '</a>'
 	};
 	if (s62aCases.SiteAddress) {
 		const address = addressToViewModel(s62aCases.SiteAddress);
--- a/apps/portal/src/app/views/applications/view/view-model.ts
+++ b/apps/portal/src/app/views/applications/view/view-model.ts
@@ -17,6 +17,8 @@ import type { Prisma } from '@pins/crowndev-database/src/client/client.ts';
 import type { BaseDevelopmentView } from '@pins/crowndev-lib/util/shared-view-model.ts';
 import { baseDevelopmentSelect, isInquiry, isHearing } from '@pins/crowndev-lib/util/shared-view-model.ts';
 
+import { insertWbr } from '@pins/crowndev-lib/util/string.ts';
+
 export const formatDate = (date: Date | null | undefined, options?: { format?: string }): string =>
 	formatDateForDisplay(date as Date | undefined, options);
 
@@ -158,7 +160,8 @@ export function crownDevelopmentToViewModel(
 }
 
 export interface CrownDevelopmentCaseListView extends BaseDevelopmentView {
-	applicantOrganisations: string[];
+	applicantOrganisations: string;
+	referenceLink: string;
 	withdrawnDate: string | undefined;
 }
 
@@ -178,14 +181,22 @@ export type CrownDevelopmentCaseListPayload = Prisma.CrownDevelopmentGetPayload<
  */
 export function applicationListViewFormattingFunction(crownDevelopment: CrownDevelopmentCaseListPayload) {
 	const extendedFields: Omit<CrownDevelopmentCaseListView, keyof BaseDevelopmentView | 'developmentContactEmail'> = {
-		applicantOrganisations: [],
+		applicantOrganisations: '',
+		referenceLink:
+			'<a class="govuk-link" href="/applications/' +
+			crownDevelopment.id +
+			'/application-information">' +
+			insertWbr(crownDevelopment.reference) +
+			'</a>',
 		withdrawnDate: undefined
 	};
 
 	if (crownDevelopment.Organisations && crownDevelopment.Organisations.length > 0) {
 		extendedFields.applicantOrganisations = crownDevelopment.Organisations.filter(
 			(organisation) => organisation.role === ORGANISATION_ROLES_ID.APPLICANT
-		).map((item) => item.Organisation.name);
+		)
+			.map((item) => item.Organisation.name)
+			.join(', ');
 	}
 
 	const withdrawnDateFormatted = formatDate(crownDevelopment.withdrawnDate, { format: 'd MMMM yyyy' });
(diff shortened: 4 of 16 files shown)


apps/s62a-portal/src/app/config.ts
```diff
@@ -3,7 +3,7 @@ import { loadEnvFile } from 'node:process';
 import type { BaseConfig } from '@pins/crowndev-lib/app/config-types.d.ts';
 
 export interface Config extends BaseConfig {
-	cacheControl: {
+	staticCacheControl: {
```
pinssbrett: Whilst it's good to rename to `staticCacheControl` to anticipate any dynamic caching on s62a-portal, your migration has only been partial. There are still some `cacheControl` references around. I think you also need to think about what should exist on `BaseConfig` and what should be on the app-level configs, as there appears to be some duplication.

Also, should this maybe be a separate PR? I don't see any link between it and the case list table macros.

apps/manage/src/app/views/cases/list/view.njk
```diff
@@ -4,11 +4,17 @@
...
+{% from "tables/tables.njk" import baseCaseListTable, baseManageHeaders, baseManageColumns %}
 
 {% block pageTitle %}
     Case list - All Cases - Manage a Crown Development Application
 {% endblock %}
 
+{% set tableAttributes = 
+    { "data-module": "moj-sortable-table" 
+    }
+ %}
```
pinssbrett: As this is the only usage of attributes, might it be better for `baseCaseListTable` to have an `isSortable` param instead? It would make the behaviour more clear if we don't anticipate having to customise the attributes beyond this

packages/lib/views/tables/tables.njk
```diff
@@ -0,0 +1,123 @@
...
+                            text: development[column.field]
+                        })
+                }), rowData) %}
+
+            {% else %}
+                {% set rowData = (rowData.push({
+                    text: development[column.field]
+                }), rowData) %}
+            
+            {% endif %}
```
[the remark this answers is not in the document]
pinssbrett: I can't see any specific concatenation, so I'm guessing nunjucks is doing it automatically? You do need spaces because it's just a comma at the moment. But unlike the copilot comment above I think this is best handled before the data gets to the template, so the template is always expecting a string value 

Image

packages/lib/views/tables/tables.njk
```diff
@@ -0,0 +1,123 @@
...
+            {% if column.type == "link" %}
+                {% set rowData = (rowData.push({
+                    html: "<a class='govuk-link' href='" +
+                        column.hrefPrefix +
+                        development.id +
+                        column.hrefSuffix +
+                        "'>" +
+                        development[column.field] +
+                        "</a>"
+                }), rowData) %}
```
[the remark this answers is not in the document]
pinssbrett: Before this refactor `insertWbr` was used for references only. I think this is another thing that should be moved into typescript so less conditional formatting is required in the template.

packages/lib/views/tables/tables.njk
```diff
@@ -0,0 +1,123 @@
...
+    {text: "Local planning authority"}
+] %}
+
+{% set basePortalColumns = [
+    {
+        type: "link",
+        field: "reference",
+        hrefPrefix: "/applications/",
+        hrefSuffix: "/application-information"
+    },
```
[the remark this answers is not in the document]
pinssbrett: See other comment - probably move to TS

pinssbrett (changes requested):
Looking good! I have a few comments where I think a wider change in moving formatting to before the template is required, which goes beyond what copilot has suggested. Also the cache control thing maybe needs splitting off

apps/s62a-portal/src/app/config.ts
andrewlilley-solirius: Agreed - Definitely a casualty of the scope creep work cut in rebasing, and needs a more careful eye on my part.

Will stash the cache work and keep it for another PR.

apps/manage/src/app/views/cases/list/view.njk
andrewlilley-solirius: That would certainly make sense. I wasn't sure if there might be future need for customisation, so coded the macro with the insertable tableAttributes - will change that since param seems preferable

packages/lib/views/tables/tables.njk
```diff
@@ -0,0 +1,130 @@
+{% from "govuk/components/table/macro.njk" import govukTable %}
+{% from "govuk/components/tag/macro.njk" import govukTag %}
+
+{% macro insertWbr(inputString) %}
+     {{ inputString | escape | replace("/", "/<wbr>") | safe }}
+ {% endmacro %}
+
```
pinssbrett: No longer used
```suggestion
```

packages/lib/util/shared-view-model.ts
```diff
@@ -23,6 +23,10 @@ export function isHearing(procedureId: string | null): boolean {
 	return procedureId === APPLICATION_PROCEDURE_ID.HEARING;
 }
 
+export function insertWbr(str: string): string {
+	return str.replace(/\//g, '/<wbr>');
+}
+
```
pinssbrett: Might be better in `util/string.ts`

packages/lib/views/tables/tables.njk
```diff
@@ -0,0 +1,130 @@
...
+                }), rowData) %}
+
+            {% else %}
+                 {% set fieldValue = development[column.field] %}
+                 {% set hasJoin = fieldValue and fieldValue.join %}
+                 
+                 {% set rowData = (rowData.push({
+                     text: (fieldValue | join(", ") if hasJoin else fieldValue),
+                     classes: column.classes
+                 }), rowData) %}
```
pinssbrett: Is this needed? Arrays should now be handled before they get to the template
```suggestion               
                 {% set rowData = (rowData.push({
                     text: development[column.field],
                     classes: column.classes
                 }), rowData) %}
```

apps/s62a-portal/src/app/views/list/view-model.ts
```diff
@@ -24,14 +25,17 @@ export type S62ADevelopmentPayload = Prisma.CrownDevelopmentGetPayload<{
  */
 export function s62aViewFormattingFunction(s62aDevelopment: S62ADevelopmentPayload) {
 	const extendedFields: Omit<S62ADevelopmentView, keyof BaseDevelopmentView | 'developmentContactEmail'> = {
-		applicantOrganisations: [],
-		withdrawnDate: s62aDevelopment.withdrawnDate
+		applicantOrganisations: '',
+		withdrawnDate: s62aDevelopment.withdrawnDate,
+		referenceLinks: insertWbr('/applications/' + s62aDevelopment.id + '/application-information')
```
pinssbrett: this is a URL, so you shouldn't be inserting `<wbr>` here. build the full link here including the `<a>` tag and reference (with wbr inserted) so you don't have to have the `if column.type == "link" ` check in the template. Make sure it works for both Crown and S62A

also `referenceLink` should be singular as there is one link for one case here

Image

packages/lib/views/tables/tables.njk
```diff
@@ -0,0 +1,130 @@
...
+        {% set rowData = [] %}
+        
+        {% for column in columns %}
+            {% if column.type == "link" %}
+                {# Safely escape the URL variables and the display text #}
+                {% set safeId = development.id | escape %}
+                {% set safeText = development[column.field] | escape %}
+                {% set rowData = (rowData.push({
+                    html: "<a class='format-link' href='" + development[column.linkField] + "'>" + development[column.field] + "</a>"
+                }), rowData) %}
```
pinssbrett: See my comment in the view model - build this link in the Typescript instead to avoid this special case. Make sure you use the `govuk-link` class and not `format-link`

pinssbrett (changes requested):
A few more simplifications and bug fixes needed, but nearly there!

apps/s62a-portal/src/app/views/list/view-model.ts
```diff
@@ -24,14 +25,22 @@ export type S62ADevelopmentPayload = Prisma.CrownDevelopmentGetPayload<{
...
-		applicantOrganisations: [],
-		withdrawnDate: s62aDevelopment.withdrawnDate
+		applicantOrganisations: '',
+		withdrawnDate: s62aDevelopment.withdrawnDate,
+		referenceLink:
+			'<a class="govuk-link" href="/applications/' +
+			s62aDevelopment.id +
+			'/application-information"><wbr>' +
+			s62aDevelopment.reference +
+			'</a>'
```
pinssbrett, line 36: - You need to apply the `insertWbr` function to the reference, not put a manual one before the reference
- I think this would be better as its own function

```suggestion
		referenceLink:
			'<a class="govuk-link" href="/applications/' +
			s62aDevelopment.id +
			'/application-information">' +
			insertWbr(s62aDevelopment.reference) +
			'</a>'
```

packages/lib/util/string.ts
```diff
@@ -82,3 +82,10 @@ const HTML_ESCAPES: Record<string, string> = {
 export function escapeHtml(value: string): string {
 	return value.replace(/[&<>"']/g, (c) => HTML_ESCAPES[c]);
 }
+
+/**
+ * Adds word break opportunities for pathing
+ */
+export function insertWbr(str: string): string {
```
pinssbrett: I should have picked this up on the last review, sorry - this could do with some tests

apps/s62a-portal/src/app/views/list/view-model.test.ts
```diff
@@ -80,8 +80,10 @@ describe('view-model', () => {
 			assert.deepStrictEqual(result, {
 				id: 'id-1',
 				reference: 'reference-id-1',
+				referenceLink:
+					'<a class="govuk-link" href="/applications/id-1/application-information"><wbr>reference-id-1</a>',
```
pinssbrett: Use a more realistic value for reference
```suggestion
reference: 'REF/2025/001',
				referenceLink:
					'<a class="govuk-link" href="/applications/id-1/application-information">REF/<wbr>2025/<wbr>001</a>',
```

apps/s62a-portal/src/app/views/list/controller.ts
```diff
@@ -72,6 +72,8 @@ export function buildCaseListPage(service: S62APortalService): AsyncRequestHandl
 			totalItems: totalS62aDevelopments
 		};
 
+		console.log(s62aDevelopmentsViewModels);
+
```
pinssbrett: unintentional console.log
```suggestion
```

apps/manage/src/app/views/cases/list/controller.js
```diff
@@ -51,9 +52,13 @@ export function crownDevelopmentToViewModel(crownDevelopment) {
 		id: crownDevelopment.id,
 		reference: crownDevelopment.reference,
 		lpaName: crownDevelopment.Lpa?.name,
-		status: crownDevelopment.Status?.displayName,
+		status: crownDevelopment.Status?.displayName
```
pinssbrett: Could also be abstracted to a function for easier reading

apps/manage/src/app/views/cases/list/controller.js
```diff
@@ -51,9 +52,13 @@ export function crownDevelopmentToViewModel(crownDevelopment) {
...
+			? '<span class="govuk-tag">' + escapeHtml(crownDevelopment.Status?.displayName) + '</span>'
+			: undefined,
 		type: crownDevelopment.Type?.displayName,
-		location: ''
+		location: '',
+		referenceLink:
```
pinssbrett, line 58: Needs `insertWbr` (see other comment)

pinssbrett (dismissed):
It's coming together well - just needs some fixes around reference formatting with wbr

I also noticed that you recently merged the s62a manage cases view. Do you need to rebase on main and make sure that uses this new refactored template too?

packages/lib/views/tables/tables.njk
```diff
@@ -0,0 +1,99 @@
...
+
+{% macro baseCaseListTable(developments, headers, columns, isSortable) %}
+    {% set rows = [] %}
+    
+    {% for development in developments %}
+        {% set rowData = [] %}
+        
+        {% for column in columns %}
+            {% set rowData = (rowData.push({
+                html: development[column.field],
```
paymonsattar, line 12: I'm not sure ingesting these fields raw is a good idea.

Let's say if organisation name is provided here as `Smith & Sons` this would break the html here. Can you give it a test? If it doesn't break it then it should be fine

packages/lib/views/tables/tables.njk
```diff
@@ -0,0 +1,99 @@
...
+        attributes: {
+            "aria-sort": "none"
+        }
+    }
+] %}
+
+{% set baseManageColumns = [
+    {
+        field: "referenceLink",
+        sortValueField: "reference"
```
paymonsattar: I don't think this is implemented, I'd suggest either implementing it or removing... it was used here but seems it was removed https://github.com/Planning-Inspectorate/crown-developments/pull/1042/changes#diff-b8cd0d9c5d12783315c6cc05eeaf04dec07a78e1b49872a856e2af46fb8eb455L65

apps/portal/src/app/views/applications/view/view-model.ts
```diff
@@ -178,14 +179,22 @@ export type CrownDevelopmentCaseListPayload = Prisma.CrownDevelopmentGetPayload<
  */
 export function applicationListViewFormattingFunction(crownDevelopment: CrownDevelopmentCaseListPayload) {
 	const extendedFields: Omit<CrownDevelopmentCaseListView, keyof BaseDevelopmentView | 'developmentContactEmail'> = {
-		applicantOrganisations: [],
+		applicantOrganisations: '',
+		referenceLink:
+			'<a class="govuk-link" href="/applications/' +
+			crownDevelopment.id +
+			'/application-information"><wbr>' +
```
paymonsattar: I think this falls under what @pinssbrett mentions in some of his other comments. 

"You need to apply the insertWbr function to the reference, not put a manual one before the reference" 

Also, you have a test here that asserts the broken output ([here](https://github.com/Planning-Inspectorate/crown-developments/pull/1042/changes#diff-9c10a6776ec1b3f981831c6adbb32bee0c2bd3bfb728fcba916104a8d256ca79R647)) - so that will also need changing

apps/s62a-portal/src/app/views/list/view.njk
```diff
@@ -1,5 +1,6 @@
 {% extends "views/layouts/main.njk" %}
 
+{% from "tables/tables.njk" import baseCaseListTable, basePortalHeaders, basePortalColumns %}
 {% from "govuk/components/table/macro.njk" import govukTable %}
```
paymonsattar: `govukTable`, `govukButton` and `govukTag` are all unused imports

apps/manage/src/app/views/cases/list/view.njk
```diff
@@ -4,6 +4,7 @@
 {% from "govuk/components/table/macro.njk" import govukTable %}
```
paymonsattar: Same as my other comment about unused imports, for `.njk` files the linter won't flag unused imports so you'll have to be wary of these yourself

packages/lib/util/string.ts
```diff
@@ -82,3 +82,21 @@ const HTML_ESCAPES: Record<string, string> = {
...
+export function insertWbr(str?: string): string {
+	return typeof str === 'string' ? str.replace(/\//g, '/<wbr>') : '';
+}
+
+/**
+ * Formats case table status tags
+ */
+export function formatStatusTag(displayName: string | null | undefined) {
+	if (displayName === null) return null;
+	if (displayName === undefined) return undefined;
```
paymonsattar: Little confused by this logic - I'm not sure why we need to distinguish between null and undefined and check for both? I would recommend an approach like below (you'd need to narrow the type and update tests accordingly. 

Unless there is something I am missing here


```
	export function formatStatusTag(displayName?: string | null): string | undefined {
     	if (!displayName) return undefined;

  	    return '<span class="govuk-tag">' + escapeHtml(displayName) + '</span>';
     }
```

paymonsattar (changes requested):

apps/manage/src/app/views/s62a/cases/list/view-model.ts
```diff
@@ -11,7 +12,7 @@ export interface S62ACaseView {
 	applicantOrganisations: string[];
 	location: string;
 	type?: string;
-	status?: string;
+	status?: string | undefined | null;
```
paymonsattar: This type is too wide, you no longer get null back from `formatStatusTag` so you can tighten this

paymonsattar (changes requested):

paymonsattar (approved):
LGTM

PR #1044, opened 2026-06-19
feat(manage): make update a case more generic

paymonsattar:
## Describe your changes
This takes the single large editsToDatabaseUpdates function, which previously handled every field for a case in one place, and split it into three pieces: 

- small shared core (applySharedUpdates, kept in `view-model.ts`) that handles the fields behaving identically for every case type - currently just description, expected submission date, type, and LPA 
- Crown-specific part moved out into a new `crown-edits.ts` file (covering everything particular to Crown, like procedures, events, organisations, fees, and statuses), 
-`s62a-edits.ts` file holding the one field unique to S62A as a demonstration. editsToDatabaseUpdates now takes a "which case type am I?" function as an argument, running the shared core first and then whichever case-type part it's handed

`update-case.ts` was updated to pass in the Crown part since it serves the Crown route.This is done so shared fields live in one place rather than being copy-pasted and fixed separately per case type, so each case type only touches its own fields and Crown and S62A can't interfere with each other.

S62A is intentionally minimal for now because the plan is to promote fields into the shared core one at a time as you confirm each behaves identically across case types.

https://pins-ds.atlassian.net/browse/PEAS-12


apps/manage/src/app/views/cases/view/update-case.ts
```diff
@@ -122,9 +123,18 @@ export function buildUpdateCase(service: ManageService, clearAnswer: boolean = f
...
 			// inside a single interactive transaction.
 			await db.$transaction(async (tx) => {
+				// Only fetch the organisation graph when the edit needs it: organisation/contact writes,
+				// or agent removal (hasAgent -> false), which may cascade organisation/contact deletes.
+				const includeOrganisations = hasOrganisationWriteEdits(toSave) || toSave.hasAgent === false;
+				const include = (
+					includeOrganisations
+						? CROWN_DEVELOPMENT_VIEW_INCLUDE
+						: { ...CROWN_DEVELOPMENT_VIEW_INCLUDE, Organisations: undefined }
+				) as typeof CROWN_DEVELOPMENT_VIEW_INCLUDE;
```
pinssbrett: Having more strictly typed includes means that (via the `CrownDevelopmentPayload` derived type), the data passed to `crownDevelopmentToViewModel` is clear. The type assertion here is hiding the real structure. Would it be better to define both includes (with and without Organisations) as separate constants and derive types from them both? Then `crownDevelopmentToViewModel` could accept the union of these and it would be clear what different data types are being passed.

apps/manage/src/app/views/cases/view/update-case.ts
```diff
@@ -122,9 +123,18 @@ export function buildUpdateCase(service: ManageService, clearAnswer: boolean = f
 			// Build a deterministic write-plan (no shared nested Organisations writes) and execute it
 			// inside a single interactive transaction.
 			await db.$transaction(async (tx) => {
+				// Only fetch the organisation graph when the edit needs it: organisation/contact writes,
+				// or agent removal (hasAgent -> false), which may cascade organisation/contact deletes.
+				const includeOrganisations = hasOrganisationWriteEdits(toSave) || toSave.hasAgent === false;
```
pinssbrett, line 128: I wonder if this logic is similar enough to that on line 171 (i.e. 'needs Organisation data') that it should be centralised somewhere?

apps/manage/src/app/views/cases/view/view-model.ts
```diff
@@ -30,6 +17,8 @@ export interface CrownDevelopmentViewModel {
 	containsDistressingContent: YesNo;
 	typeId: string;
 	subTypeId?: string;
+	// ── S62A PLACEHOLDER
+	applicationClassificationId?: string;
```
pinssbrett, line 22: Adding S62A fields to the Crown view model is going to create a big combined type that doesn't reflect either. I think each case type should extend a shared view model of common fields (same as in PEAS-9). Or is that in the scope of another ticket?

apps/manage/src/app/views/cases/view/view-model.ts
```diff
@@ -594,210 +583,69 @@ function assignDirectUpdate<K extends DirectUpdateField>(
...
+ */
+export function editsToDatabaseUpdates(
+	edits: CrownDevelopmentSaveModel,
+	viewModel: CrownDevelopmentViewModel,
+	applyCaseTypeUpdates: ApplyCaseTypeUpdates,
+	options: EditsToDatabaseUpdatesOptions = {}
+): Prisma.CrownDevelopmentUpdateInput {
+	const updateInput: Prisma.CrownDevelopmentUpdateInput = {};
+	applySharedUpdates(updateInput, edits);
+	applyCaseTypeUpdates(updateInput, edits, viewModel, options);
```
pinssbrett, line 622: Is there a benefit to keeping the existing mutation behaviour here? If these functions returned the updates instead it might be easier to unit test them

apps/manage/src/app/views/cases/view/s62a-edits.ts
pinssbrett, line 1: Should this function have tests?

pinssbrett (changes requested):
Update handling for the different service types looks great. I have a few questions around implementation (the shared view model one being the biggest)

apps/manage/src/app/views/cases/view/update-case.ts
paymonsattar: Good shout here, I've updated

apps/manage/src/app/views/cases/view/update-case.ts
paymonsattar: Good question - they share the hasOrganisationWriteEdits check, but the agent halves differ on purpose (line 130 is intentionally broader because it runs before dbViewModel exists, whereas line 171 can check the stored yes to no transition), so I'd keep them as two sites and just add clarifying comments rather than centralise behind a helper whose name would have to hide that pre-fetch/post-fetch difference.

apps/manage/src/app/views/cases/view/view-model.ts
paymonsattar: I think this should be for another ticket - currently I'm not sure exactly what new fields will be added. I do agree with you that this could become bloated once we have some more s62A fields - but this can be addressed in a future ticket. What do you think?

apps/manage/src/app/views/cases/view/view-model.ts
paymonsattar: Yeah, good shout on the testing side - but editsToDatabaseUpdates already returns its result so we can test it directly, and I'd keep the internal mutation since the case-type updaters deliberately overwrite shared keys like Stage and procedureNotificationDate; returning and merging would trade that clear last-write-wins ordering for shallow-merge behaviour that'd quietly break the first time two functions touch the same nested relation.

apps/manage/src/app/views/cases/view/s62a-edits.ts
paymonsattar: True, I've added tests for this and crown edit

apps/manage/src/app/views/cases/view/crown-edits.test.ts
```diff
@@ -0,0 +1,534 @@
...
+			assert.strictEqual(updates.ApplicantContact?.update?.orgName, 'Applicant One');
+			assert.strictEqual(updates.ApplicantContact?.update?.email, 'applicant@example.com');
+			assert.strictEqual(updates.ApplicantContact?.update?.Address?.upsert?.where?.id, 'address-id-1');
+			assert.strictEqual(updates.ApplicantContact?.update?.Address?.upsert?.create?.line1, 'Street');
+			assert.strictEqual(updates.ApplicantContact?.update?.Address?.upsert?.update?.line1, 'Street');
+			assert.strictEqual(updates.ApplicantContact?.update?.Address?.upsert?.create?.townCity, 'City');
+			assert.strictEqual(updates.ApplicantContact?.update?.Address?.upsert?.update?.townCity, 'City');
+			assert.strictEqual(updates.ApplicantContact?.update?.Address?.upsert?.create?.postcode, 'PSC01D');
+			assert.strictEqual(updates.ApplicantContact?.update?.Address?.upsert?.update?.postcode, 'PSC01D');
+		});
```
pinssbrett: FYI this functionality is being [simultaneously removed](https://github.com/Planning-Inspectorate/crown-developments/pull/1047)

apps/manage/src/app/views/cases/view/update-case.ts
pinssbrett: This cast is still present and telling the compiler downstream that organisations are in `crownDevelopment` when they're not. Removing the cast will highlight where the issue now lies. Could we use function overloads to make the return type of the Prisma call conditional on `includeOrganisations`? Alternatively, would it be cleaner to split this into two separate fetch paths where each has a properly typed result, even if it means some duplication? Is there a reason we're not using your new `CROWN_DEVELOPMENT_VIEW_INCLUDE_WITHOUT_ORGS` constant here instead of manually setting Organisations: undefined?

apps/manage/src/app/views/cases/view/view-model.ts
pinssbrett: Okay, but if so I think we still need to get it sorted early in this build, before we start adding fields. If for a different ticket then include ticket reference in a TODO

pinssbrett (changes requested):
I think there's still a bit of type clarity needed on the with/without organisations flow. Otherwise we're looking good!

pinssbrett (approved):
Looks good to me!

PR #1046, opened 2026-06-22
feat: adds major or non-major question

oscargay:
## Describe your changes
- Adds a new conditional question after `application` answer, asking whether the application is `major or non-major`

[PEAS-55](https://pins-ds.atlassian.net/browse/PEAS-55)

## Implementation
Screenshot 2026-06-22 at 13 26 32
Screenshot 2026-06-22 at 13 26 35
Screenshot 2026-06-22 at 13 26 43

[PEAS-55]: https://pins-ds.atlassian.net/browse/PEAS-55?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/create-a-case/questions.ts
+++ b/apps/manage/src/app/views/s62a/cases/create-a-case/questions.ts
@@ -4,18 +4,30 @@ import {
 	COMPONENT_TYPES,
 	RequiredValidator
 } from '@planning-inspectorate/dynamic-forms';
-import { PRE_APPLICATION_OR_APPLICATIONS } from '@pins/crowndev-database/src/seed/s62a/data-static.ts';
+import {
+	MAJOR_OR_NON_MAJORS,
+	PRE_APPLICATION_OR_APPLICATIONS
+} from '@pins/crowndev-database/src/seed/s62a/data-static.ts';
 
 export function getQuestions() {
 	const questions = {
-		preApplicationOrApplication: {
+		applicationStage: {
 			type: COMPONENT_TYPES.RADIO,
-			title: 'Pre-application or application?',
+			title: 'Application stage',
 			question: 'Is this a pre-application or an application?',
-			fieldName: 'preApplicationOrApplication',
+			fieldName: 'applicationStage',
 			url: 'pre-application-or-application',
 			validators: [new RequiredValidator('Select whether this is a pre-application or an application')],
 			options: PRE_APPLICATION_OR_APPLICATIONS.map((t) => ({ text: t.displayName, value: t.id }))
+		},
+		applicationClassification: {
+			type: COMPONENT_TYPES.RADIO,
+			title: 'Application classification',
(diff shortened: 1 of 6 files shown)


apps/manage/src/app/views/s62a/cases/create-a-case/questions.ts
```diff
@@ -16,6 +19,15 @@ export function getQuestions() {
 			url: 'pre-application-or-application',
 			validators: [new RequiredValidator('Select whether this is a pre-application or an application')],
 			options: PRE_APPLICATION_OR_APPLICATIONS.map((t) => ({ text: t.displayName, value: t.id }))
+		},
+		majorOrNonMajor: {
+			type: COMPONENT_TYPES.RADIO,
+			title: 'Major or non-major?',
```
pinssbrett: I know this is the title on the ticket, but functionally this is used for the check your answers page, where referencing the design it should probably be 'Application category'.
This also applies to the question above, which should be 'Application stage'.
I'm sure this will get picked up down the line but might be good to sort now.

```suggestion
			title: 'Application category',
```

apps/manage/src/app/views/s62a/cases/create-a-case/questions.ts
```diff
@@ -16,6 +19,15 @@ export function getQuestions() {
...
+			question: 'Is this a major or non-major application?',
+			fieldName: 'majorOrNonMajor',
+			url: 'major-or-non-major',
```
pinssbrett, line 28: Considering the title change above, it might be worth double-checking the required URL too, which isn't on the ticket

pinssbrett (changes requested):
Just a couple of naming things to check

apps/manage/src/app/views/s62a/cases/create-a-case/questions.ts
oscargay: @pinssbrett - yep, you're right! I had planned to do it as per ticket and then handle the titling within the specific CYA ticket itself at the end, but happy to address now

apps/manage/src/app/views/s62a/cases/create-a-case/questions.ts
oscargay: Hmm, yes I did run the URLs by the BAs, but will check now with the title change

oscargay:
@pinssbrett - back with you! I spoke to the BAs and they have updated tickets to reflect CYA, there were some ones that were not quite right in the mural, so have updated. I have:
1) Updated the titles to reflect
2) URL kept the same
3) I have updated the fieldNames and object keys as well to match as I think that's more developer friendly than the other, even though we don't quite have data model yet.

Thanks!

pinssbrett (approved):
Nice, looks good!

PR #1048, opened 2026-06-22
feat(peas-153): build out components

edward-mitchell-gov:
Linked to Jira ticket: https://pins-ds.atlassian.net/browse/PEAS-153

Added reusable components for common pages. E.g date and address entry. 
These methods will need to be put into use and refined further before turning them into a installer package.


playwright/page-components/address.component.ts
```diff
@@ -0,0 +1,179 @@
...
+
+type AddressErrorConfig = {
+	message: string;
+	inlineId: string;
+	inputId: AddressFieldKey;
+};
+
+const addressErrorMap: Record<AddressErrorType, AddressErrorConfig> = {
+	line1TooLong: {
+		message: 'Address line 1 must be 250 characters or less',
```
AlexMoulin: Would it be worth having the character limit of error messages a value that can be updated? Something like:
 const ADDRESS_LINE_1_MAX_CHAR = 250.

playwright/page-components/date.component.ts
AlexMoulin, line 1: Does this work with dates that aren't 'received dates'?

playwright/page-utilities/generate.utility.ts
```diff
@@ -0,0 +1,376 @@
...
+}
+
+/**
+ * Generates realistic text for validation testing:
+ * - mostly real words
+ * - occasionally inserts safe special-character words
+ * - always returns the exact requested length
+ * - never ends with a space
+ */
+export function generateRandomString(length: number): string {
```
AlexMoulin: Does this create a risk of tests occasionally failing because a randomString contains specialWords only 10% of the time? Like say parenthesis cause a failure, this will only happen 1 in 120 times

playwright/page-utilities/generate.utility.ts
```diff
@@ -0,0 +1,376 @@
...
+/**
+ * Generates a varied linked or related case reference.
+ */
+export function generateCaseReference(type: 'related' | 'linked'): string {
+	const prefixes =
+		type === 'related'
+			? ['REL', 'related', 'RELATED', 'case', 'app', 'plan']
+			: ['LNK', 'linked', 'LINKED', 'case', 'ref', 'way'];
+
+	const caseTypes = [
```
AlexMoulin: I don't think you need to generate case references this way. Crown doesn't have multiple caseTypes in that sense. References are CROWN/YYYY/{7 digit int} and I think the plan is for S62a to be S62A/YYYY/{7 digit int}

AlexMoulin (changes requested):
Looks good! There are some considerations I have added but nothing major

playwright/page-utilities/generate.utility.ts
edward-mitchell-gov: I've removed this function completely, you are right that it will be unrelated to crown. 
It will be easier to create a new function if and when required to populate date for references.

playwright/page-utilities/generate.utility.ts
edward-mitchell-gov: Good points. i've updated the function to have a default of safe word strings. There is now an option to include special characters in the generated string and also the ability to add special examples when required.

playwright/page-components/address.component.ts
edward-mitchell-gov: Good point. The address component currently uses the same validation limits across the apps I’ve worked on, so I don’t think this needs to be configurable per test/page. But defining the shared limits as constants would make the mapping clearer and easier to update if the component validation changes later. I’ll add constants for the address and postcode limits.

playwright/page-components/date.component.ts
edward-mitchell-gov: Good point. I’ve made the date utility configurable so it no longer assumes every date is a received date. It now accepts a fieldPrefix and dateName, defaulting to the existing received date values, so pages using fields like startDate can reuse the same validation logic with the correct error messages.

AlexMoulin (approved):
LGTM

PR #973, opened 2026-05-26
Feat/peas 9 case list on s62a portal

andrewlilley-solirius:
## Describe your changes
Skeleton for the S62A case list for future development. Currently using crown database as placeholder. Pared down version of the **manage** app case list without add case or search functionality.

## Issue ticket number and link
PEAS-9
https://pins-ds.atlassian.net/browse/PEAS-9

PR #1049, opened 2026-06-22
Feat/peas 124 type of application

oscargay:
## Describe your changes
- Adds 2 questions for type of application
- Questions are basically identical but have variations in title and question depending on pre-application vs application
- NB: due to more strict typing from dynamic forms I needed to make changes to some more core crown types

https://pins-ds.atlassian.net/browse/PEAS-124

## Implementation
Screenshot 2026-06-22 at 14 52 48
Screenshot 2026-06-22 at 14 52 51
Screenshot 2026-06-22 at 14 52 59
Screenshot 2026-06-22 at 14 53 03

### DIFF
--- a/apps/manage/src/app/views/cases/view/view-model.ts
+++ b/apps/manage/src/app/views/cases/view/view-model.ts
@@ -24,7 +24,7 @@ import type { CrownDevelopmentPayload } from './payload-contracts.ts';
 type ProcedurePrefix = 'inquiry' | 'hearing' | 'writtenReps';
 type ProcedureId = (typeof APPLICATION_PROCEDURE_ID)[keyof typeof APPLICATION_PROCEDURE_ID];
 
-export interface CrownDevelopmentViewModel {
+export type CrownDevelopmentViewModel = {
 	reference: string;
 	description: string;
 	containsDistressingContent: YesNo;
@@ -169,7 +169,7 @@ export interface CrownDevelopmentViewModel {
 	hasCostsApplications?: YesNo;
 	costsApplicationsComment?: string;
 	siteVisitDate?: Date;
-}
+};
 
 export type CrownDevelopmentViewModelFields = keyof CrownDevelopmentViewModel;
 
--- a/apps/manage/src/app/views/cases/view/controller.ts
+++ b/apps/manage/src/app/views/cases/view/controller.ts
@@ -25,6 +25,7 @@ import type { SharePointDrive } from '@pins/crowndev-sharepoint/src/sharepoint/d
 import type { Response, Request, Handler, NextFunction } from 'express';
 import type { ErrorSummaryItem } from '@pins/crowndev-lib/util/types.ts';
 import type { ManageService } from '#service';
+import type { CrownJourneyResponse } from '../../../../types/express-locals.d.ts';
 
 /**
  * Get the journey answers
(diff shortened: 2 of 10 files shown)


paymonsattar (dismissed):
LGTM!

apps/manage/src/app/views/cases/view/view-model.ts
```diff
@@ -24,7 +24,7 @@ import type { CrownDevelopmentPayload } from './payload-contracts.ts';
 type ProcedurePrefix = 'inquiry' | 'hearing' | 'writtenReps';
 type ProcedureId = (typeof APPLICATION_PROCEDURE_ID)[keyof typeof APPLICATION_PROCEDURE_ID];
 
-export interface CrownDevelopmentViewModel {
+export type CrownDevelopmentViewModel = {
```
pinssbrett, line 27: Similar to [on the portal](https://github.com/Planning-Inspectorate/crown-developments/pull/973/changes#diff-a7d2050f469500c195fae013c49ed8c2921fd1b00876e1916f527e1a06598428R22), the view model here is going to likely want to extend a shared generic view with s62A specific fields. This has also come up on the [update-case ticket](https://github.com/Planning-Inspectorate/crown-developments/pull/1044#discussion_r3451621555). I don't think there's a need to do it as part of this but you might want to keep the interface (since it will be extending) and add a TODO?

packages/database/src/seed/s62a/data-static.ts
```diff
@@ -29,3 +29,35 @@ export const MAJOR_OR_NON_MAJORS = [
...
 	}
 ];
+
+export const APPLICATION_TYPE_ID = Object.freeze({
+	PLANNING_PERMISSION: 'planning-permission',
+	OUTLINE_SOME_MATTERS_RESERVED: 'outline-some-matters-reserved',
+	OUTLINE_ALL_MATTERS_RESERVED: 'outline-all-matters-reserved',
+	APPROVAL_OF_RESERVED_MATTERS: 'approval-of-reserved-matters',
+	PLANNING_PERMISSION_AND_LBC: 'planning-permission-and-lbc'
+} as const);
```
pinssbrett: I just want to sense check defining these entirely separately from the crown data if they are the same. Should shared fields come from a shared file? Also these represent the same things as the Crown options but the values are slightly different - should they be aligned?

pinssbrett (commented):
Looks good, just a couple of broad implementation questions

apps/manage/src/app/views/cases/view/view-model.ts
oscargay: @pinssbrett - thanks, yeah the only reason I did this was because of some TS errors that started cropping up where because it was an interface it couldn't satisfy the generic typing of `[k: string]: unknown` of the `JourneyAnswers` typing from dynamic forms. I could just leave it as an interface and use a ts-ignore for now?

packages/database/src/seed/s62a/data-static.ts
oscargay: Ah good spot, I think I should probably just be using the data from the crown file for this. Because we aren't doing anything with the data model yet I just picked "sensible" keys but if there are already ones then I will use those. It might be worth doing a bit of a restructure at some point of crown vs s62a vs shared in a separate ticket in the future?

apps/manage/src/app/views/cases/view/view-model.ts
pinssbrett: I'm fine to keep the change to from interface to type and we can use union types if abstracting.

packages/database/src/seed/s62a/data-static.ts
pinssbrett: Agreed - probably best to get that abstraction ticket in sooner rather than later, though, so we don't end up diverging too much in the meantime

pinssbrett (approved):
Looks good!

PR #1057, opened 2026-06-26
feat: adds agent contact information add to list

oscargay:
## Describe your changes
- Adds the agent contact information add to list flow

PEAS-135

## Implementation
Screenshot 2026-06-29 at 11 44 45
Screenshot 2026-06-29 at 11 44 52
Screenshot 2026-06-29 at 11 44 57
Screenshot 2026-06-29 at 11 45 09
Screenshot 2026-06-29 at 11 45 16
Screenshot 2026-06-29 at 11 45 34
Screenshot 2026-06-29 at 11 45 38

### DIFF
--- a/packages/lib/forms/custom-components/manage-list/question.js
+++ b/packages/lib/forms/custom-components/manage-list/question.js
@@ -14,6 +14,7 @@ import nunjucks from 'nunjucks';
  * 	isAllowedEmpty?: boolean;
  * 	confirmRemoveButtonText?: string;
  * 	removalPrompt?: string;
+ * 	forceInitialAdd?: boolean;
  * }} CustomManageListQuestionProps
  */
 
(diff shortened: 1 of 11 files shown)


packages/lib/forms/custom-components/manage-list/question.js
```diff
@@ -14,6 +14,7 @@ import nunjucks from 'nunjucks';
  * 	isAllowedEmpty?: boolean;
  * 	confirmRemoveButtonText?: string;
  * 	removalPrompt?: string;
+ * 	forceInitialAdd?: boolean;
```
AlexMoulin, line 17: What does forceInitialAdd do? I don't understand how it differs from the normal addButton. Would it be worth adding a comment here?

AlexMoulin (changes requested):

packages/lib/forms/custom-components/manage-list/question.js
oscargay: @AlexMoulin - `forceInitialAdd` basically forces the user to add at least one thing by only showing the add button. Adding a comment is a good idea but i actually refactor the wording and structure of this in a child ticket when there is another flag added, so I think it's not worth doing in this ticket but probably that one. Is that okay?

AlexMoulin (approved):

PR #1072, opened 2026-07-03
fix(azure): update Azure AI Language client authentication to use ManagedIdentityCredential

AlexMoulin:
## Describe your changes
This pull request updates how the Azure AI Language client is authenticated in the `ManageService` class. The main change is switching from `DefaultAzureCredential` to `ManagedIdentityCredential` for authentication as DefaultAzureCredentials would prioritise EnvAuthentication and fail for Azure Language AI.

**Azure Authentication Updates:**

* Updated the import in `service.js` to include `ManagedIdentityCredential` from `@azure/identity` alongside `DefaultAzureCredential`.
* Changed the initialization of `TextAnalyticsClient` to use `ManagedIdentityCredential` instead of `DefaultAzureCredential` when setting up the Azure AI Language client.

image

https://pins-ds.atlassian.net/browse/CROWN-1682


RobBroadhead30 (approved):
lgtm

PR #1077, opened 2026-07-07
feat/peas 111 manage s62a application list case table 

andrewlilley-solirius:
## Describe your changes
Application List ('All cases') page which serves as the landing page of the Manage s62A application service. 

- Lists all s62A cases in a sortable table
- Has both an empty state (no cases created) and a populated state
- Reached from the ‘All cases’ tab in the header and also the ‘Return to your case list’ CTA on the final page of the Create a Case journey

Tickets dependent on this branch:
 - PEAS-201 (Back office case list search)
 - PEAS-242 (Back office header)

[https://pins-ds.atlassian.net/browse/PEAS-111](url)

[CROWN-1328]: https://pins-ds.atlassian.net/browse/CROWN-1328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PEAS-242]: https://pins-ds.atlassian.net/browse/PEAS-242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/index.ts
+++ b/apps/manage/src/app/views/s62a/cases/index.ts
@@ -1,13 +1,17 @@
-import { Router as createRouter } from 'express';
-import { createRoutes as createCreateACaseRoutes } from './create-a-case/index.ts';
-import { createRoutes as createCaseRoutes } from './view/index.ts';
 import type { ManageService } from '#service';
+import { Router as createRouter } from 'express';
+import { asyncHandler } from '@pins/crowndev-lib/util/async-handler.ts';
+import { createRoutes as createCaseRoutes } from './view/index.ts';
+import { createRoutes as createCreateACaseRoutes } from './create-a-case/index.ts';
+import { buildCaseListPage } from './list/controller.ts';
 
 export function createRoutes(service: ManageService) {
 	const router = createRouter({ mergeParams: true });
 	const createACaseRoutes = createCreateACaseRoutes(service);
 	const caseRoutes = createCaseRoutes(service);
+	const homePageController = buildCaseListPage(service);
 
+	router.get('/', asyncHandler(homePageController));
 	router.use('/create-a-case', createACaseRoutes);
 	router.use('/:id', caseRoutes);
 
--- a/apps/manage/src/app/views/s62a/cases/list/controller.test.ts
+++ b/apps/manage/src/app/views/s62a/cases/list/controller.test.ts
@@ -1,0 +1,269 @@
+import { describe, it, mock } from 'node:test';
+import assert from 'node:assert';
+import { buildCaseListPage } from './controller.ts';
+import { mockLogger } from '@pins/crowndev-lib/testing/mock-logger.js';
+import type { ManageService } from '#service';
+import type { BaseLogger } from 'pino';
+import type { Request, Response } from 'express';
+
+const PAGINATION_TEST_CASES = [
+	{
+		name: 'should generate 1 page for 25 total items requested from page 1',
+		totalItems: 25,
+		itemsPerPage: 25,
+		requestedPage: 1,
+		expected: { totalPages: 1, resultsStartNumber: 1, resultsEndNumber: 25 }
+	},
+	{
+		name: 'should generate 2 pages and return the first 25 items for 50 total items requested from page 1',
+		totalItems: 50,
+		itemsPerPage: 25,
+		requestedPage: 1,
+		expected: { totalPages: 2, resultsStartNumber: 1, resultsEndNumber: 25 }
+	},
+	{
+		name: 'should generate 2 pages and return the last 25 items for 50 total items requested from page 2',
+		totalItems: 50,
+		itemsPerPage: 25,
+		requestedPage: 2,
+		expected: { totalPages: 2, resultsStartNumber: 26, resultsEndNumber: 50 }
+	},
+	{
+		name: 'should generate 3 pages and return the 2nd set of 25 items for 60 total items requested from page 2',
+		totalItems: 60,
+		itemsPerPage: 25,
+		requestedPage: 2,
+		expected: { totalPages: 3, resultsStartNumber: 26, resultsEndNumber: 50 }
+	},
+	{
+		name: 'should generate 3 pages and return the final 10 items for 60 total items requested from page 3 (partial page)',
+		totalItems: 60,
+		itemsPerPage: 25,
+		requestedPage: 3,
+		expected: { totalPages: 3, resultsStartNumber: 51, resultsEndNumber: 60 }
+	},
+	{
+		name: 'should generate 4 pages and return the 3rd set of 25 items for 100 total items requested from page 3',
+		totalItems: 100,
+		itemsPerPage: 25,
+		requestedPage: 3,
+		expected: { totalPages: 4, resultsStartNumber: 51, resultsEndNumber: 75 }
+	},
+	{
+		name: 'should generate 4 pages and return the last 25 items for 100 total items requested from page 4',
+		totalItems: 100,
+		itemsPerPage: 25,
+		requestedPage: 4,
+		expected: { totalPages: 4, resultsStartNumber: 76, resultsEndNumber: 100 }
+	}
+];
+
+/**
+ * Creates Mock cases, currently used for the pagination tests
+ * as they need a lot of data.
+ */
+const createMockCases = (count: number) => {
+	const cases = [];
+	for (let i = 1; i <= count; i++) {
+		cases.push({
+			id: `id-${i}`,
+			reference: `S62A/${i}`,
+			ApplicantContact: {
+				orgName: `Applicant ${i}`
+			},
+			Lpa: {
+				name: 'Test Council'
+			},
+			Stage: {
+				displayName: 'Test Stage'
+			},
+			Type: {
+				displayName: 'Test Type'
+			}
+		});
+	}
+	return cases;
+};
+
+describe('case list', () => {
+	describe('list published cases', () => {
+		const nunjucks = {
+			render: mock.fn((view, data) => 'mocked render' + view + data)
+		};
+		it('should render page without error', async () => {
+			const mockReq = {
+				params: {
+					id: 'some-id'
+				}
+			} as unknown as Request;
+
+			const mockResData = {
+				render: mock.fn((view, data) => nunjucks.render(view, data))
+			};
+			const mockRes = mockResData as unknown as Response;
+
+			const mockDb = {
+				s62aCase: {
+					findMany: mock.fn(() => [
+						{
+							id: 'id-1',
+							reference: 'S62A/1',
+							ApplicantContact: {
+								orgName: 'John Smith'
+							},
+							Lpa: {
+								name: 'System Test Borough Council'
+							},
+							Stage: {
+								displayName: 'Inquiry'
+							},
+							Type: {
+								displayName: 'Planning permission'
+							}
+						},
+						{
+							id: 'id-2',
+							reference: 'S62A/2',
+							ApplicantContact: {
+								orgName: 'Dave James'
+							},
+							Lpa: {
+								name: 'System Test Borough Council'
+							},
+							Stage: {
+								displayName: 'Hearing'
+							},
+							Type: {
+								displayName: 'Outline planning permission with some matters reserved'
+							}
+						}
+					]),
+					count: mock.fn(() => 2)
+				}
+			} as unknown;
+
+			const applicationList = buildCaseListPage({
+				db: mockDb,
+				logger: mockLogger() as unknown as BaseLogger,
+				s62aDevContactInfo: { email: 's62a.dev@gov.uk' }
+			} as unknown) as any;
+
+			await assert.doesNotReject(() => applicationList(mockReq, mockRes));
+
+			assert.strictEqual(mockResData.render.mock.callCount(), 1);
+			assert.strictEqual(mockResData.render.mock.calls[0].arguments.length, 2);
+			assert.strictEqual(mockResData.render.mock.calls[0].arguments[0], './views/s62a/cases/list/view.njk');
+			assert.strictEqual(mockResData.render.mock.calls[0].arguments[1].pageTitle, 'Manage Section 62A applications');
+			assert.strictEqual(mockResData.render.mock.calls[0].arguments[1].s62aCasesViewModels.length, 2);
+		});
+		it('should render page without error when no crown dev cases returned', async () => {
+			const mockReq = {
+				params: {
+					id: 'some-id'
+				}
+			} as unknown as Request;
+			const mockResData = {
+				render: mock.fn((view, data) => nunjucks.render(view, data))
+			};
+			const mockRes = mockResData as unknown as Response;
+
+			const mockDb = {
+				s62aCase: {
+					findMany: mock.fn(() => []),
+					count: mock.fn(() => 0)
+				}
+			} as unknown;
+
+			const service = {
+				db: mockDb,
+				logger: mockLogger() as unknown as BaseLogger,
+				s62aDevContactInfo: { email: 's62a.dev@gov.uk' }
+			} as unknown as ManageService;
+
+			const applicationList = buildCaseListPage(service);
+
+			await assert.doesNotReject(() => applicationList(mockReq, mockRes));
+
+			assert.strictEqual(mockResData.render.mock.callCount(), 1);
+			assert.strictEqual(mockResData.render.mock.calls[0].arguments.length, 2);
+			assert.strictEqual(mockResData.render.mock.calls[0].arguments[0], './views/s62a/cases/list/view.njk');
+			assert.deepStrictEqual(mockResData.render.mock.calls[0].arguments[1], {
+				pageTitle: 'Manage Section 62A applications',
+				s62aCasesViewModels: [],
+				baseUrl: '/s62a/cases',
+				containerClasses: 'pins-container-wide',
+				currentUrl: undefined,
+				queryParams: undefined,
+				paginationParams: {
+					pageNumber: 1,
+					resultsEndNumber: 0,
+					resultsStartNumber: 0,
+					selectedItemsPerPage: 25,
+					totalItems: 0,
+					totalPages: 0
+				}
+			});
+		});
+		describe('Pagination permutations', () => {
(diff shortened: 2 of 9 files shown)


apps/manage/src/app/views/s62a/cases/list/controller.test.ts
```diff
@@ -0,0 +1,273 @@
...
+		itemsPerPage: 25,
+		requestedPage: 4,
+		expected: { totalPages: 4, resultsStartNumber: 76, resultsEndNumber: 100 }
+	}
+];
+
+/**
+ * Creates Mock cases, currently used for the pagination tests
+ * as they need a lot of data.
+ * @param {number} count
```
oscargay: Can remove JSDoc parameter definitions when in TypeScript as it only serves to potentially cause confusion!

apps/manage/src/app/views/s62a/cases/list/controller.test.ts
```diff
@@ -0,0 +1,273 @@
...
+			Type: {
+				displayName: 'Test Type'
+			}
+		});
+	}
+	return cases;
+};
+
+describe('case list', () => {
+	describe('list published cases', () => {
```
oscargay, line 89: NIT: I am guessing that this is from `portal` - I think the manage BO this list will show all cases, even if they aren't published yet?

apps/manage/src/app/views/s62a/cases/list/controller.test.ts
```diff
@@ -0,0 +1,273 @@
...
+				}
+			} as unknown as Request;
+
+			const mockResData = {
+				render: mock.fn((view, data) => nunjucks.render(view, data))
+			};
+			const mockRes = mockResData as unknown as Response;
+
+			const mockDb = {
+				crownDevelopment: {
```
oscargay: Hmm, this has got me thinking... The schema changes for S62A are currently in review, so it probably makes sense to wait for that, rebase on top of `main` and update all this to use real S62A cases? What do you think?

apps/manage/src/app/views/s62a/cases/list/controller.ts
```diff
@@ -0,0 +1,91 @@
...
+		} catch (error) {
+			wrapPrismaError({
+				error,
+				logger,
+				message: 'fetching database entries',
+				logParams: {}
+			});
+		}
+
+		logger.info(`Crown development list page: ${s62aDevelopments.length} case(s) fetched`);
```
oscargay: Logger should say "S62A list page..." or something to that effect !

apps/manage/src/app/views/s62a/cases/list/controller.ts
```diff
@@ -0,0 +1,91 @@
...
+		);
+
+		const { totalPages, resultsStartNumber, resultsEndNumber } = getPageData(
+			totalItems,
+			selectedItemsPerPage,
+			pageSize,
+			pageNumber
+		);
+
+		interface PaginationParams {
```
oscargay: ~Nice interface, feels like it might be good to have this in some sort of Util somewhere as I assume it will want to be reused all over the app~

Above still stands, but just noticed that Rob has a PR up for standardising pagination across the app. Might be something that needs to merged and rebased for this PR to avoid immediate divergence / conflicts!

apps/manage/src/app/views/s62a/cases/list/view.njk
```diff
@@ -0,0 +1,145 @@
+{% extends "views/layouts/main.njk" %}
+
+{% from "govuk/components/button/macro.njk" import govukButton %}
```
oscargay: I don't think all of these imports are used, might be worth double checking!

apps/manage/src/app/views/s62a/cases/list/view.njk
```diff
@@ -0,0 +1,145 @@
...
+{% from "govuk/components/table/macro.njk" import govukTable %}
+{% from "govuk/components/tag/macro.njk" import govukTag %}
+{% import "pagination/pagination.njk" as pagination %}
+{% import "pagination/pagination-options.njk" as paginationOptions %}
+{% from "views/layouts/components/header.njk" import header %}
+{% from "search-bar/search-bar.njk" import searchBar %}
+
+{% block pageTitle %}
+    Case list - All Cases - Manage a Crown Development Application
```
oscargay: Title here is still referring to Crown Devs!

apps/manage/src/app/views/s62a/cases/list/view.njk
```diff
@@ -0,0 +1,145 @@
...
+                        classes: "govuk-input govuk-input--width-30"
+                    },
+                    button: {
+                        text: "Search",
+                        classes: "govuk-button"
+                    }
+                }, searchValue) }}
+            </form>
+        </div>
+        <div class="govuk-grid-column-one-third pins-dashboard-box">
```
oscargay: Is there a wireframe for this page? I know there are plans to update the header soon to have the create button in it, but it might be fine to do this like Crown for now and update it when we get there?

apps/manage/src/app/views/s62a/cases/list/view.njk
```diff
@@ -0,0 +1,145 @@
...
+                    </h2>
+
+            </div>
+        </div>
+    </div>
+
+    <div id="pagination" class="govuk-body">
+        {% if s62aDevelopmentsViewModels|length == 0 %}
+            <p class="govuk-body">No cases have been created yet</p>
+        {% elif s62aDevelopmentsViewModels|length <= 25 %}
```
oscargay: The pagination should handle this case by default, I think in other places in the app we don't do this. If there are fewer than 25 cases the pagination controls won't show, but it will still say Showing X to Y of Y cases which I think is useful!

apps/manage/src/app/views/s62a/cases/list/view.njk
```diff
@@ -0,0 +1,145 @@
...
+            {{ casesTable(s62aDevelopmentsViewModels) }}
+       {% endif %}
+    </div>
+
+    {# if pageNumber and totalPages values are both 1 then pagination options not shown #}
+    {{ pagination.renderPagination(paginationParams.pageNumber, paginationParams.totalPages, baseUrl, queryParams) }}
+
+{% endblock %}
+
+{% macro casesTable(developmentCases) %}
```
oscargay: Syntax here is still referring to crown developments, worth updating just so its simpler down the line!

apps/manage/src/app/views/s62a/cases/list/view.njk
```diff
@@ -0,0 +1,145 @@
...
+                html: govukTag({text: development.status})
+            }
+        ]), developments) %}
+    {% endfor %}
+
+    {% set headers = [
+        {
+            text: 'Reference',
+            attributes: {
+            "aria-sort": "descending"
```
oscargay: NIT: spacing can get weird in nunjucks files, but these aria-sort should be indented one more!

apps/manage/src/app/views/s62a/cases/view/view-model.ts
```diff
@@ -0,0 +1,240 @@
...
+	truncatedReadMoreCommentLink
+} from '@pins/crowndev-lib/util/questions.ts';
+import {
+	isWithdrawnOrExpired,
+	isExpired,
+	type ApplicationPublishStatus
+} from '@pins/crowndev-lib/util/applications.ts';
+import type { Prisma } from '@pins/crowndev-database/src/client/client.ts';
+
+//TODO - REPLACE THIS WITH PEAS-9 TICKET FUNCTIONS INSTEAD
```
oscargay: 👍  - I think most of this file is basically going to be removed to start. We'll make a basic model that is enough to satiate the create a case journey, and then slowly add to it as we do tickets for the case details page. Most of this won't be needed especially to start, especially towards the bottom with the functions surrounding applications and stuff

apps/manage/src/app/views/s62a/cases/index.ts
```diff
@@ -1,10 +1,15 @@
+import type { ManageService } from '#service';
 import { Router as createRouter } from 'express';
+import { asyncHandler } from '@pins/crowndev-lib/util/async-handler.ts';
 import { createRoutes as createCreateACaseRoutes } from './create-a-case/index.ts';
+import { buildCaseListPage } from './list/controller.ts';
 
-export function createRoutes() {
+export function createRoutes(service: ManageService) {
```
oscargay, line 8: Apologies you're going to have a few conflicts probably as I have done something similar for my work!

apps/manage/src/app/config.js
```diff
@@ -233,6 +234,9 @@ export function loadConfig() {
 		},
 		// the static directory to serve assets from (images, css, etc..)
 		staticDir: buildConfig.staticDir,
+		s62aDevContactInfo: {
+			email: S62A_DEV_CONTACT_EMAIL
```
oscargay: You can probably just leave this out for now, unless it has been specced? Otherwise will need to add it to the terraform and stuff for it to even appear, so doesn't add much!

apps/manage/src/app/config-types.d.ts
```diff
@@ -57,4 +57,7 @@ interface Config {
 		caseTemplateId?: string; // Id for template folder (new case template folder structure)
 	};
 	staticDir: string;
+	s62aDevContactInfo: {
```
oscargay: As below, I think for now we can just leave this out?

apps/portal/src/app/views/applications/view/have-your-say/save.js
```diff
@@ -1,7 +1,7 @@
 import { isValidUuidFormat } from '@pins/crowndev-lib/util/uuid.ts';
 import { notFoundHandler } from '@pins/crowndev-lib/middleware/errors.ts';
 import { crownDevelopmentToViewModel } from '../view-model.ts';
-import { fetchPublishedApplication } from '#util/applications.ts';
+import { fetchPublishedApplication } from '@pins/crowndev-lib/util/applications.ts';
```
oscargay, line 4: Was this a part of your abstraction work? I'm a bit unclear why this is inside of this PR, It looks like it should have been done in something else?

apps/portal/jsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
 	"extends": "../../jsconfig.json",
-	"include": ["**/*.js"],
+	"include": ["**/*.js", "**/*.d.ts"],
```
oscargay: Why was this change needed?

oscargay (changes requested):
Great work Andrew! Some comments to consider, but overall looking good. I think we will have to wait to merge and get the S62A model in first to make sure this has those changes implemented.

apps/manage/src/app/views/s62a/cases/list/controller.test.ts
andrewlilley-solirius: That would make a lot of sense - making a note to check when that's merged

apps/portal/src/app/views/applications/view/have-your-say/save.js
andrewlilley-solirius: That's definitely snuck in, changing that now. I believe this happened after rebasing the 17-into-6 PEAS-9 commits and stashing excess view-model refactoring across the portal app.

Made a note on the Pre-Commit/PR Checklist to watch out for this as I also re-introduced a number of older bugs into PEAS-9 as a result.

apps/portal/jsconfig.json
andrewlilley-solirius: That's crept in as well. Removing that now

apps/manage/src/app/views/s62a/cases/list/controller.ts
```diff
@@ -0,0 +1,57 @@
...
+	const { db, logger } = service;
+	return async (req, res) => {
+		const { pageSize, skipSize } = getPaginationParams(req);
+
+		let s62aDevelopments: S62ADevelopmentPayload[] = [];
+		let totalItems: number = 0;
+
+		try {
+			[s62aDevelopments, totalItems] = await Promise.all([
+				db.crownDevelopment.findMany({
```
oscargay: Looks like this is still using `crownDevelopments`? There's no reason we need to use them anymore, as we have a S62A model now in the database! We should change it now to avoid having to undo everything down the line? Thanks!

apps/manage/src/app/views/s62a/cases/list/view.njk
```diff
@@ -0,0 +1,111 @@
...
+        {% endif %}
+    </div>
+
+    {# if pageNumber and totalPages values are both 1 then pagination options not shown #}
+    {{ pagination.renderPagination(paginationParams.pageNumber, paginationParams.totalPages, baseUrl, queryParams) }}
+
+{% endblock %}
+
+{% macro casesTable(s62aCases) %}
+    {% set developments = [] %}
```
oscargay: NIT: they aren't developments for S62A, I think we can just call them cases or s62aCaseRows or something?

apps/manage/src/app/views/s62a/cases/view/view-model.ts
```diff
@@ -120,3 +123,79 @@ export function s62aCaseToViewModel(dbCase: S62aCaseDbModel): S62aCaseViewModel
 
 	return viewModel;
 }
+
+export interface S62ADevelopmentView extends BaseDevelopmentView {
```
oscargay: Hmmm, I'm not really 100% sure what's going on here, I am guessing it's because you are using a crown development model? But we shouldn't have / need any of these values for S62A right now, it should be the bare minimum view needed for the list view, which should only be a few columns I think. Unless I am misunderstanding something? Feel free to shoot me a message!

apps/manage/src/app/views/s62a/cases/view/view-model.ts
```diff
@@ -120,3 +123,79 @@ export function s62aCaseToViewModel(dbCase: S62aCaseDbModel): S62aCaseViewModel
...
+} satisfies Prisma.CrownDevelopmentSelect;
+
+export type S62ADevelopmentPayload = Prisma.CrownDevelopmentGetPayload<{
+	select: typeof s62aDevelopmentSelect;
+}>;
+
+/**
+ * S62A view model formatter, formatting extended fields from S62A Development View
+ *
+ * @param s62aDevelopment - The main db query input
```
oscargay: NIT: we don't need any JSDocs when using typescript!

apps/manage/.env.example
```diff
@@ -22,6 +22,7 @@ SHAREPOINT_CASE_TEMPLATE_ID='templateId'
 
 #GovNotify
 GOV_NOTIFY_DISABLED=false
+S62A_DEV_CONTACT_EMAIL=an-email
```
oscargay: Can remove this I think

apps/manage/src/app/service.js
```diff
@@ -85,6 +85,7 @@ export class ManageService {
 		this.getEntraClient = buildInitEntraClient(!config.auth.disabled, entraGroupCache);
 		this.notifyClient = initGovNotify(config.govNotify, logger);
 		this.blobStoreClient = initBlobStore(config.blobStore, logger);
+		this.s62aDevContactInfo = config.s62aDevContactInfo;
```
oscargay: I don't think this done anything now right?

apps/manage/src/app/views/s62a/cases/view/view-model.ts
```diff
@@ -3,6 +3,9 @@ import { SITE_AREA_UNIT_ID } from '@pins/crowndev-database/src/seed/s62a/data-st
 import { addressToViewModel } from '@pins/crowndev-lib/util/address.ts';
 import type { YesNo } from '@pins/crowndev-lib/util/types.ts';
 import { type Address, booleanToYesNoValue } from '@planning-inspectorate/dynamic-forms';
+import { ORGANISATION_ROLES_ID } from '@pins/crowndev-database/src/seed/data-static.ts';
```
oscargay: The stuff you have added is for the `s62a/cases/list/*` directory, you should I think add a `view-model.ts` file there with its own tests rather than using the file designed for the case details view page. It should be relatively small, it just needs to take the data and map it to what's needed for the columns, as you have done at the bottom of this file

oscargay (changes requested):
Nice work! Back with you!

There are a few comments I have, but the main one is that we don't need to use crown developments anymore as we have our own s62a schema, by swapping to using that I think then some of the confusion / comments I had regarding the view model typing might go away, as currently I am a little confused as to why that view model is pulling so many unused / irrelevant fields? But as I said, I could be missing something! Either way, good job so far :)

apps/manage/src/app/views/s62a/cases/list/view-model.ts
```diff
@@ -0,0 +1,67 @@
...
+	const fields = {
+		id: s62aCases.id,
+		reference: s62aCases.reference,
+		lpaName: s62aCases.Lpa?.name,
+		status: s62aCases.S62aStatus?.displayName ?? undefined,
+		type: s62aCases.Type?.displayName ?? undefined,
+		applicantOrganisations:
+			s62aCases.S62aToApplicants?.filter(
+				(relation) =>
+					relation.roleId === ORGANISATION_ROLES_ID.APPLICANT && typeof relation.Organisation?.name === 'string'
```
oscargay, line 46: In our app it could be an organisation or an individual, but I think this is fine for now and if we want to change it down the line it's easy enough to do!

apps/manage/src/app/views/s62a/cases/list/view.njk
```diff
@@ -0,0 +1,111 @@
...
+{% from "views/layouts/components/header.njk" import header %}
+{% from "search-bar/search-bar.njk" import searchBar %}
+
+{% block pageTitle %}
+    Case list - All Cases - Manage a Section 62A Development Application
+{% endblock %}
+
+{% block header %}
+	{{ header('Manage a Section 62A Application', baseUrl ) }}
+{% endblock %}
```
oscargay, line 16: This is fine for now I think, but we are going to want a S62A main.njk file at some point that we can extend so we don't have to do `block header` everywhere. But I know we have a ticket to sort the header out coming up so that should be fine to do then!

oscargay (approved):
LGTM! nice work Andrew

PR #1078, opened 2026-07-07
Feat/peas 201 s62a manage application list search case reference

andrewlilley-solirius:
## Describe your changes
Adding case search to the S62A manage application list ‘all cases’ page to allow back-office users to search for a case by reference. 

Depends on PEAS-111 (Back office case list)

[https://pins-ds.atlassian.net/browse/PEAS-201](url)

[CROWN-1328]: https://pins-ds.atlassian.net/browse/CROWN-1328?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PEAS-242]: https://pins-ds.atlassian.net/browse/PEAS-242?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/list/controller.ts
+++ b/apps/manage/src/app/views/s62a/cases/list/controller.ts
@@ -2,6 +2,7 @@ import type { ManageService } from '#service';
 import type { AsyncRequestHandler } from '@pins/crowndev-lib/util/async-handler.ts';
 import { wrapPrismaError } from '@pins/crowndev-lib/util/database.ts';
 import { createPaginationParams, getPaginationParams } from '@pins/crowndev-lib/views/pagination/pagination-utils.ts';
+import { normaliseSearchQuery, splitStringQueries, createWhereClause } from '@pins/crowndev-lib/util/search-queries.js';
 
 import { s62aToViewModel, s62aCaseSelect } from './view-model.ts';
 import type { S62ACaseView, S62ACasePayload } from './view-model.ts';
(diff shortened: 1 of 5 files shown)


apps/manage/src/app/views/s62a/cases/list/controller.ts
```diff
@@ -5,18 +5,34 @@ import { getPageData, getPaginationParams } from '@pins/crowndev-lib/views/pagin
...
 
+		const rawSearchCriteria = req.query?.searchCriteria;
+
+		const searchCriteriaParam = Array.isArray(rawSearchCriteria)
+			? typeof rawSearchCriteria[0] === 'string'
+				? rawSearchCriteria[0]
+				: undefined
+			: typeof rawSearchCriteria === 'string'
+				? rawSearchCriteria
+				: undefined;
```
oscargay: NIT: same logic but might be nice to move this into a function and split into 2 variables for easy scanning. Could go into the generic `search-queries.js` util file

```
const normaliseSearchQuery = (query: string | array) => {
    const normalisedString = Array.isArray(query) ? query[0] : query;

    return typeof normalisedString === 'string' ? normalisedString : undefined;
}

```

Then other users of the search query can reuse it !

oscargay (changes requested):
Looks good! I have 1 note, but apart from that all nice, will not approve because it needs the parent merged + some other bits to be in main. Nice work though :)

packages/lib/util/search-queries.js
```diff
@@ -12,6 +12,17 @@ export function splitStringQueries(query) {
 		.filter(Boolean);
 }
 
+/**
+ * Normalise a query input into a string or undefined
+ * @param {string | QueryString.ParsedQs | (string | QueryString.ParsedQs)[] | undefined} query - Query string, array of strings, or parsed query object
+ * @returns {string | undefined} - Normalised string, or undefined
+ */
+export const normaliseSearchQuery = (query) => {
```
oscargay, line 20: Super small one, but might be worth adding a super simple unit test for this inside of the test file? Just checking the various outcomes that could be passed in!

oscargay (approved):
LGTM!

PR #1079, opened 2026-07-07
feat: adds the skeleton s62a schema and cya page

oscargay:
## Describe your changes
- Creates initial schema for S62A model, only does enough needed for create a case, with expectations of adding as and when we need new fields in the frontend.

https://pins-ds.atlassian.net/browse/PEAS-177

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/create-a-case/s62a-case-mapper.ts
+++ b/apps/manage/src/app/views/s62a/cases/create-a-case/s62a-case-mapper.ts
@@ -1,0 +1,385 @@
+import { Prisma } from '@pins/crowndev-database/src/client/client.ts';
+import { ORGANISATION_ROLES_ID } from '@pins/crowndev-database/src/seed/data-static.ts';
+import {
+	APPLICANT_TYPE_ID,
+	S62A_STATUS_ID,
+	SITE_AREA_UNIT_ID
+} from '@pins/crowndev-database/src/seed/s62a/data-static.ts';
+import type { YesNo } from '@pins/crowndev-lib/util/types.ts';
+import { yesNoToBoolean } from '@planning-inspectorate/dynamic-forms';
+
+export interface AddressAnswers {
+	addressLine1?: string;
+	addressLine2?: string;
+	townCity?: string;
+	county?: string;
+	postcode?: string;
+}
+
+export interface ApplicantOrganisationAnswer {
+	id?: string;
+	organisationName: string;
+	organisationAddress?: AddressAnswers;
+}
+
+export interface ApplicantContactAnswer {
+	id?: string;
+	applicantFirstName?: string;
+	applicantLastName?: string;
+	applicantContactEmail?: string;
+	applicantContactTelephoneNumber?: string;
+	applicantContactOrganisation?: string;
+}
+
+export interface AgentContactAnswer {
+	id?: string;
+	agentFirstName?: string;
+	agentLastName?: string;
+	agentContactEmail?: string;
+	agentContactTelephoneNumber?: string;
+}
+
+export interface CreateCaseAnswers {
+	// LPA Details
+	lpaId: string;
+	lpaFirstName?: string;
+	lpaLastName?: string;
+	lpaEmailAddress?: string;
+	lpaPhoneNumber?: string;
+	hasSecondaryLpa?: YesNo;
+	secondaryLpaId?: string;
+	secondaryLpaFirstName?: string;
+	secondaryLpaLastName?: string;
+	secondaryLpaEmailAddress?: string;
+	secondaryLpaPhoneNumber?: string;
+
+	// Agent
+	hasAgent: YesNo;
+	agentName?: string;
+	agentAddress?: AddressAnswers;
+	manageAgentContactDetails?: AgentContactAnswer[];
+
+	// Applicant
+	applicantType?: 'organisation' | 'individual';
+	manageApplicantOrganisations?: ApplicantOrganisationAnswer[];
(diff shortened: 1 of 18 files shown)


packages/database/src/schema.prisma
```diff
@@ -159,6 +159,115 @@ model CrownDevelopment {
...
+  contactId String?  @db.UniqueIdentifier
+  Contact   Contact? @relation(fields: [contactId], references: [id], onUpdate: NoAction, onDelete: NoAction)
+
+  roleId String
+  Role   S62aToApplicantRole @relation(fields: [roleId], references: [id], onUpdate: NoAction, onDelete: NoAction)
+
+  @@index([s62aId])
+  @@index([organisationId])
+  @@index([contactId])
+}
```
pinssbrett, line 269: Is there any enforcement of the exclusive arc structure on the database side or are we relying on application-level validation? If the latter, there's a risk of introducing data errors. I wonder if we need a `type` column and some kind of `CHECK` constraint to enforce this. Or should it be separate join tables?

packages/database/src/schema.prisma
```diff
@@ -159,6 +159,115 @@ model CrownDevelopment {
...
+  SiteAddress            Address?      @relation(fields: [siteAddressId], references: [id], onUpdate: NoAction, onDelete: NoAction)
+  siteEasting            Int?
+  siteNorthing           Int?
+  siteAreaInHectares     Decimal?
+  siteAreaOriginalUnitId String?
+  SiteAreaOriginalUnit   SiteAreaUnit? @relation(fields: [siteAreaOriginalUnitId], references: [id])
+
+  /// Core dates
+  notificationSubmittedDate DateTime?
+  expectedSubmissionDate    DateTime?
```
pinssbrett: Should this be non-nullable, since the field is required?
```suggestion
  expectedSubmissionDate    DateTime
```

apps/manage/src/app/views/s62a/cases/create-a-case/s62a-case-mapper.ts
```diff
@@ -0,0 +1,386 @@
...
+			input.SiteAddress = {
+				create: this.toAddressInput(this.answers.siteAddress)
+			};
+		}
+
+		if (this.answers.siteAreaHectares) {
+			input.siteAreaInHectares = parseFloat(this.answers.siteAreaHectares);
+			input.SiteAreaOriginalUnit = { connect: { id: SITE_AREA_UNIT_ID.HECTARES } };
+		} else if (this.answers.siteAreaSquareMetres) {
+			input.siteAreaInHectares = parseFloat(this.answers.siteAreaSquareMetres) / 10000;
```
pinssbrett: Is there a risk of introducing floating point errors here?

apps/manage/src/app/views/s62a/cases/create-a-case/s62a-case-mapper.ts
```diff
@@ -0,0 +1,386 @@
...
+	id?: string;
+	agentFirstName?: string;
+	agentLastName?: string;
+	agentContactEmail?: string;
+	agentContactTelephoneNumber?: string;
+}
+
+export interface CreateCaseAnswers {
+	// LPA Details
+	lpaId?: string;
```
pinssbrett: Some fields are required, so should these not be optional parameters? e.g. `lpaId`, `applicationType`, `hasAgent`
```suggestion
	lpaId: string;
```

apps/manage/src/app/views/s62a/cases/create-a-case/s62a-case-mapper.ts
```diff
@@ -0,0 +1,386 @@
...
+	lpaPhoneNumber?: string;
+	hasSecondaryLpa?: string;
+	secondaryLpaId?: string;
+	secondaryLpaFirstName?: string;
+	secondaryLpaLastName?: string;
+	secondaryLpaEmailAddress?: string;
+	secondaryLpaPhoneNumber?: string;
+
+	// Agent
+	hasAgent?: string;
```
pinssbrett: For these boolean fields stored as strings we have the `YesNo` object so the data type is more clear (`'@pins/crowndev-lib/util/types.ts'`)
```suggestion
	hasAgent: YesNo;
```

apps/manage/src/app/views/s62a/cases/create-a-case/s62a-case-mapper.ts
```diff
@@ -0,0 +1,386 @@
...
+			return null;
+		}
+
+		if (!this.answers.manageAgentContactDetails) {
+			throw new Error('Agent contacts are required when the case has an agent');
+		}
+
+		if (!this.answers.agentName) {
+			throw new Error('Agent name is required when the case has an agent');
+		}
```
pinssbrett, line 305: Do we need tests for these?

apps/manage/src/app/views/s62a/cases/create-a-case/s62a-case-mapper.ts
```diff
@@ -0,0 +1,386 @@
...
+
+	private toAddressInput(address: AddressAnswers) {
+		return {
+			line1: address.addressLine1,
+			line2: address.addressLine2,
+			townCity: address.townCity,
+			county: address.county,
+			postcode: address.postcode
+		};
+	}
```
pinssbrett, line 379: Do we need tests for this?

pinssbrett (changes requested):
Looks good and much improved compared to Crown. I have a comment around the exclusive arc part of the data structure and then some smaller things around fields.

pinssbrett (approved):
Looking good, thanks!

PR #1084, opened 2026-07-09
feat: wip case view page boilerplate

oscargay:
## Describe your changes
- Adds the boilerplate for the new S62A case details page
- Adds 2 questions, one scalar and one join to give an example of how it will work across 2 tabs too.

[PEAS-245](https://pins-ds.atlassian.net/browse/PEAS-245)

## Implementation
Screenshot 2026-07-10 at 13 50 58
Screenshot 2026-07-10 at 13 51 03

[PEAS-245]: https://pins-ds.atlassian.net/browse/PEAS-245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/view/index.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/index.ts
@@ -1,0 +1,49 @@
+import type { ManageService } from '#service';
+import { Router as createRouter } from 'express';
+import { buildGetJourneyMiddleware, buildViewCaseDetails } from './controller.ts';
+import { asyncHandler } from '@pins/crowndev-lib/util/async-handler.ts';
+import { VIEW_TAB_ID } from '@pins/crowndev-database/src/seed/s62a/data-static.ts';
+import {
+	buildGetJourneyResponseFromSession,
+	buildSave,
+	question,
+	validate,
+	validationErrorHandler
+} from '@planning-inspectorate/dynamic-forms';
+import { JOURNEY_ID } from './journey.ts';
+import { buildS62aUpdateCase } from './update-case.ts';
+
+export function createRoutes(service: ManageService) {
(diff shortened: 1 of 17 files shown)


apps/manage/src/app/views/s62a/cases/view/index.ts
```diff
@@ -0,0 +1,49 @@
...
+} from '@planning-inspectorate/dynamic-forms';
+import { JOURNEY_ID } from './journey.ts';
+import { buildS62aUpdateCase } from './update-case.ts';
+
+export function createRoutes(service: ManageService) {
+	const router = createRouter({ mergeParams: true });
+
+	const getJourney = asyncHandler(buildGetJourneyMiddleware(service));
+	const viewCaseDetails = buildViewCaseDetails();
+	const getQuestionJourney = asyncHandler(buildGetJourneyMiddleware(service));
```
paymonsattar, line 21: Just checking here - `getQuestionJourney` and `getJourney` seem to be duplicates - is this intented and if so why do we need both?

oscargay: Yeah they are identical you are right. I probably should have just left it as is and then we could've added it when we needed it. But essentially, we are going to need to have them split soon to handle the differences between what we want a question to look like on the question page vs on the details list. We do this in create-a-case (look up `isQuestionView`) but because we don't have any questions that require that check yet, there is no param, so they look identical.

As I said, I probably should have just only had 1 for now and then we could split them out into 2 when we got to the point that we needed them, as that's better dev practice, but it doesn't cause any harm as is? What do you think?

paymonsattar (approved):

PR #1088, opened 2026-07-13
feat(manage): add case notes

paymonsattar:
## Describe your changes

- Ports the case notes feature from MPESC to CROWN
- Adds the standalone notes page and the "recent 4 notes" block on the case details page, with an add-note form
- Adding notes is audited
- Notes are behind the `isCaseNotesLive` feature flag

https://pins-ds.atlassian.net/browse/PEAS-17

Screenshot 2026-07-13 at 12 46 55
Screenshot 2026-07-13 at 12 51 17


apps/manage/src/app/views/cases/view/view-model.ts
```diff
@@ -623,6 +630,41 @@ export function editsToDatabaseUpdates(
...
+			const user = members.find((member) => member.id === note.userId);
+
+			return {
+				date: formatDateForDisplay(note.createdAt, { format: 'd MMMM yyyy' }),
+				dayOfWeek: formatDateForDisplay(note.createdAt, { format: 'EEEE' }),
+				time: formatDateForDisplay(note.createdAt, { format: 'h:mmaaa' }),
+				commentText: nl2br(escape(note.comment)),
+				truncatedCommentText:
+					nl2br(escape(truncateComment(note.comment))) +
+					(shouldTruncateComment(note.comment) ? truncatedReadMoreCommentLink(readMoreHref) : ''),
```
pinssbrett: Was this supposed to truncate at a different length? Since the truncation length matches the max note length of 500 characters. Or just for future proofing any edge cases?

apps/manage/src/app/views/cases/view/controller.ts
```diff
@@ -192,6 +197,8 @@ export function buildViewCaseDetails({ db, getSharePointDrive, isApplicationUpda
 		const baseUrl = req.baseUrl;
 
 		const banner = await getBannerMessages(id, res, req, db);
+		const notes = (res.locals.caseNotes ?? []) as ReturnType<typeof mapNotes>['caseNotes'];
+		const allCaseNotesCount = (res.locals.allCaseNotesCount as number) ?? 0;
```
pinssbrett: Can you define these in `express-locals.d.ts` so our locals structure is more clear? Then no cast required

apps/manage/src/app/views/cases/case-notes/validation-middleware.ts
```diff
@@ -0,0 +1,61 @@
...
+		}
+
+		const errors = generateCaseNoteErrors(req);
+		if (errors.length) {
+			addSessionData(req, id, { updateErrors: errors }, 'cases');
+
+			return res.redirect(`/cases/${id}`);
+		}
+
+		if (next) next();
```
pinssbrett: Nitpick - shouldn't next always exist?
```suggestion
		next();
```

apps/manage/src/app/views/cases/case-notes/validation-middleware.ts
```diff
@@ -0,0 +1,61 @@
+import type { Handler } from 'express';
+import { getStringParam } from '@pins/crowndev-lib/util/params.ts';
+import { addSessionData } from '@pins/crowndev-lib/util/session.ts';
+import type { Request } from 'express';
+
+const MAX_COMMENT_LENGTH = 500;
+
+type ErrorMessage = { text: string; href: string };
```
pinssbrett: This type exists in the lib as `ErrorSummaryItem`

pinssbrett (changes requested):
Looks good, just a few small things I have questions on

apps/manage/src/app/views/cases/view/view-model.ts
paymonsattar: Sorry, this was my bad. It should be 100 but I was testing and forgot to change this back. Good spot. I'll change this back to match what MPESC was doing

apps/manage/src/app/views/cases/view/controller.ts
paymonsattar: Good shout! Updated

apps/manage/src/app/views/cases/case-notes/validation-middleware.ts
paymonsattar: Correct, I'll fix this

apps/manage/src/app/views/cases/case-notes/validation-middleware.ts
```diff
@@ -0,0 +1,60 @@
+import type { Handler } from 'express';
+import type { Request } from 'express';
+import { getStringParam } from '@pins/crowndev-lib/util/params.ts';
+import { addSessionData } from '@pins/crowndev-lib/util/session.ts';
+import type { ErrorSummaryItem } from '@pins/crowndev-lib/util/types';
```
pinssbrett: Required to fix type error in pipeline
```suggestion
import type { ErrorSummaryItem } from '@pins/crowndev-lib/util/types.ts';
```

packages/lib/util/questions.ts
```diff
@@ -3,7 +3,7 @@ import { LOCAL_PLANNING_AUTHORITIES as LOCAL_PLANNING_AUTHORITIES_DEV } from '@p
 import { LOCAL_PLANNING_AUTHORITIES as LOCAL_PLANNING_AUTHORITIES_PROD } from '@pins/crowndev-database/src/seed/data-lpa-prod.ts';
 import type { Option } from '@planning-inspectorate/dynamic-forms';
 
-const MAX_LENGTH = 500;
+export const CASE_NOTE_MAX_LENGTH = 100;
```
pinssbrett, line 6: Do you think this should this be defined where it's used instead of in lib?

apps/manage/src/app/views/cases/view/view-model.test.ts
pinssbrett, line 1: I saw you updated some tests for the other truncation to 500 chars - should we add some tests for 100 char truncation here?

pinssbrett (changes requested):
I have a couple of extra comments relating to the recent changes

packages/lib/util/questions.ts
paymonsattar: I didn't move it just yet as we need to move the case notes logic to a shared package as it will be used in both Crown and s62A. I'll do this as part of that PR

pinssbrett (dismissed):
Looks good to me!

pinssbrett (dismissed):

pinssbrett (approved):

PR #1093, opened 2026-07-14
feat: adds case details overview page

oscargay:
## Describe your changes
- Adds the overview page, including data model + tests

[PEAS-200](https://pins-ds.atlassian.net/browse/PEAS-200)

## Implementation
Screenshot 2026-07-15 at 09 29 53

[PEAS-200]: https://pins-ds.atlassian.net/browse/PEAS-200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/view/constants.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/constants.ts
@@ -1,0 +1,17 @@
+import { PRE_APPLICATION_OR_APPLICATION_ID } from '@pins/crowndev-database/src/seed/s62a/data-static.ts';
+
+/**
+ * Identical questions on the case details page can have
+ * different text dependending on whether the case
+ * is an application or pre-application.
+ */
+export const CASE_DETAILS_QUESTION_TEXT = {
+	[PRE_APPLICATION_OR_APPLICATION_ID.PRE_APPLICATION]: {
+		expectedSubmissionDateQuestion: 'What is the expected submission date for the pre-application?',
+		expectedSubmissionDateTitle: 'Expected pre-application submission date'
+	},
+	[PRE_APPLICATION_OR_APPLICATION_ID.APPLICATION]: {
+		expectedSubmissionDateQuestion: 'What is the expected submission date for the application?',
+		expectedSubmissionDateTitle: 'Expected application submission date'
+	}
+};
--- a/apps/manage/src/app/views/s62a/cases/view/questions.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/questions.ts
@@ -1,14 +1,66 @@
-import { S62A_STATUSES } from '@pins/crowndev-database/src/seed/s62a/data-static.ts';
 import {
+	APPLICATION_SUB_TYPES,
+	APPLICATION_TYPE_ID,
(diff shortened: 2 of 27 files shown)


apps/manage/src/app/views/s62a/cases/view/questions.ts
```diff
@@ -25,6 +77,258 @@ export function getQuestions() {
...
+				]
+			}
+		},
+		inspectorBand: {
+			type: COMPONENT_TYPES.RADIO,
+			title: 'Inspector band',
+			question: 'Which level of inspector is required? (optional)',
+			fieldName: 'inspectorBandId',
+			url: 'inspector-band',
+			validators: [new RequiredValidator('Select the level of inspector required')],
```
paymonsattar, line 159: The question specifies it's optional but there is a requiredValidator attached to it - is this intended?

apps/manage/src/app/views/s62a/cases/view/constants.ts
```diff
@@ -0,0 +1,12 @@
+import { PRE_APPLICATION_OR_APPLICATION_ID } from '@pins/crowndev-database/src/seed/s62a/data-static.ts';
+
+export const QUESTION_TEXT = {
```
paymonsattar: NITPICK: I notice theres now two exports with this same but different shapes - can we consider a more specific name?

apps/manage/src/app/views/s62a/cases/view/update-case.ts
```diff
@@ -23,6 +24,13 @@ export function buildS62aUpdateCase(service: ManageService): SaveDataFn {
 			return;
 		}
 
+		// Wipes answers that were indentified as being removed
+		if (clearAnswer) {
+			Object.keys(answers).forEach((key) => {
+				Object.assign(answers, { [key]: null });
```
paymonsattar, line 30: Unless I'm misunderstand - this looks like it clears all the answers not just the removed ones?

paymonsattar (changes requested):

apps/manage/src/app/views/s62a/cases/view/questions.ts
oscargay: Yeah it's intentional, it's because it is optional (you can remove it via the `remove and save` button) but because it's a radio you can't have it submit with nothing so we have a validator to stop that. I think anyway!

apps/manage/src/app/views/s62a/cases/view/update-case.ts
oscargay: You have understood correctly, it's just that this is the view case page, so you only submit 1 thing at a time. We did the same thing in MPESC and I think Crown does it too? It might be that this doesn't always support our use case but think we could just update it then as and when. What do you think

paymonsattar: Ah - I didn't think about we only submitting one at a time. Sounds good

paymonsattar (approved):

PR #1095, opened 2026-07-15
fix: commonises remove applicants function

oscargay:
## Describe your changes
- Uses the function to remove applicants when an org is removed.
- Commonises the function and tests 

[PEAS-256](https://pins-ds.atlassian.net/browse/PEAS-256)[PEAS-256](https://pins-ds.atlassian.net/browse/PEAS-256)

[PEAS-256]: https://pins-ds.atlassian.net/browse/PEAS-256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PEAS-256]: https://pins-ds.atlassian.net/browse/PEAS-256?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pinssbrett (approved):
Looks good to me! I'm wondering if the CSRF warning is a false positive since we are handling that at the app level with Lusca

oscargay:
@pinssbrett - I think you are right, will dismiss the alert, it's also not something that was touched at all in this code so confident in it not being a regression

PR #1098, opened 2026-07-15
fix: removes optional from addresses

oscargay:
## Describe your changes
- Removes a few pointless `Optional` tags from addresses

PEAS-254


pinssbrett (approved):
Looks good

PR #1101, opened 2026-07-17
fix(portal): add error summary for documents page

RobBroadhead30:
## Describe your changes

Bug fix on documents page in the portal app where the filters were displaying an inline error in the filter section but not a errorSummary. This ticket fixes the error summary so it now appears at the top when a error occurs

https://pins-ds.atlassian.net/browse/CROWN-1689

before
image

after
image

### DIFF
--- a/apps/portal/src/app/views/applications/view/documents/controller.test.ts
+++ b/apps/portal/src/app/views/applications/view/documents/controller.test.ts
@@ -857,4 +857,82 @@ describe('Date Published Filter', () => {
 		assert.strictEqual((viewData.queryParams as Record<string, string>)['publishedDateFrom-month'], '1');
 		assert.strictEqual((viewData.queryParams as Record<string, string>)['publishedDateFrom-year'], '2026');
 	});
+
+	it('should build error summary when incomplete date published from date', async () => {
+		const documents = [
+			{
+				id: '1',
+				name: 'doc.pdf',
+				file: { mimeType: 'application/pdf' },
+				createdDateTime: '2025-02-05T10:00:00Z',
+				lastModifiedDateTime: '2025-02-05T10:00:00Z',
+				size: 1000
+			}
+		];
+
+		const mockDb = createMockDb();
+		const mockSharePoint = createMockSharePoint(documents);
+		const handler = buildApplicationDocumentsPage({
+			db: mockDb,
+			logger: mockLogger(),
+			sharePointDrive: mockSharePoint
+		} as any);
+
+		const req = createMockRequest({
+			query: {
+				'publishedDateFrom-day': '',
+				'publishedDateFrom-month': '02',
+				'publishedDateFrom-year': '2025'
+			}
+		});
+
+		const res = createMockResponse();
+
+		await handler(req, res, createMockNext());
+
+		const viewData = getRenderArgs(res);
+		assert.ok(Array.isArray(viewData.errorSummary));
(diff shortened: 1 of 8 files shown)


apps/portal/src/app/views/applications/view/documents/controller.test.ts
```diff
@@ -857,4 +857,82 @@ describe('Date Published Filter', () => {
 		assert.strictEqual((viewData.queryParams as Record<string, string>)['publishedDateFrom-month'], '1');
 		assert.strictEqual((viewData.queryParams as Record<string, string>)['publishedDateFrom-year'], '2026');
 	});
+
+	it('should build error summary when incomplete date published from date', async () => {
```
AlexMoulin, line 861: I don't think the date in your test is incomplete it is invalid

apps/portal/src/app/views/applications/view/documents/controller.ts
```diff
@@ -180,7 +183,27 @@ export function buildApplicationDocumentsPage(service: PortalService): RequestHa
...
+			errorSummary
 		});
 	};
 }
+
+/**
+ * Maps error summary from filters.
+ */
+
+function mapErrorSummary(filters: FilterSection[]): Array<{ text: string; href: string }> {
```
AlexMoulin: Does a function like this not exist for written representations? Could it be changed into a util and reused?

AlexMoulin (changes requested):
Test naming needs a review. And a question about reusable code. Looks good overall though, well done!

apps/portal/src/app/views/applications/view/utils/filter-error-summary.ts
```diff
@@ -0,0 +1,67 @@
...
+type MapDateFilterErrorSummaryOptions = {
+	sectionTitle: string;
+	getHref?: (dateInput: DateFilterInput, formType: FormType) => string;
+	requireDateInputType?: boolean;
+};
+
+export function mapDateFilterErrorSummary(
+	filters: FilterSectionLike[] = [],
+	formTypeOrOptions: FormType | MapDateFilterErrorSummaryOptions = 'desktop',
+	options?: MapDateFilterErrorSummaryOptions
```
AlexMoulin: The param being one or the other here could lead to it being confusing. Your alternatives could be:
* To just change it into 1 param that uses keys to identify what everything is rather than formTypeOrOptions being one or the other.
* Require both

apps/portal/src/app/views/applications/view/utils/filter-error-summary.ts
```diff
@@ -0,0 +1,67 @@
...
+): ErrorSummaryItem[] {
+	let formType: FormType;
+	let opts: MapDateFilterErrorSummaryOptions;
+
+	if (typeof formTypeOrOptions === 'object') {
+		formType = 'desktop';
+		opts = formTypeOrOptions;
+	} else {
+		formType = formTypeOrOptions;
+		opts = options!;
```
AlexMoulin: opts is potentially null or undefined due to options being optional (options?) so asserting that it isn't could cause mask some issues.

AlexMoulin (changes requested):
Looks good, a bit more clarity and tightening the params of mapDataFilterErrorSummary and I think you will be good to go.

AlexMoulin (approved):
LGTM

PR #1102, opened 2026-07-17
Feat/peas 242 update s62a header

andrewlilley-solirius:
## Describe your changes
Update the Section 62A service header so it matches the current prototype

- Change only applies to pages under the /s62a route- Crown Development pages outside that path keep their existing header and are not affected. 
- Header on S62A pages are renamed from Manage a Crown Development Application to Manage Section 62A applications. - The layout, banding and navigation should also be brought in line with the prototype. 

Depends on PEAS-111 (Back office case list)

https://pins-ds.atlassian.net/browse/PEAS-242

### DIFF
--- a/apps/manage/src/app/views/layouts/components/core/banner.njk
+++ b/apps/manage/src/app/views/layouts/components/core/banner.njk
@@ -1,0 +1,8 @@
+{% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}
+
+{{ govukPhaseBanner({
+  tag: {
+    text: "Beta"
+  },
+  html: 'This is a new service. Help us improve it and <a class="govuk-link" href="' + config.serviceFeedbackUrl + '" target="_blank" rel="noreferrer">give your feedback (opens in new tab)</a>'
+}) }}
--- a/apps/manage/src/app/views/s62a/cases/list/view.njk
+++ b/apps/manage/src/app/views/s62a/cases/list/view.njk
@@ -2,16 +2,11 @@
 
 {% import "pagination/pagination.njk" as pagination %}
 {% import "pagination/pagination-options.njk" as paginationOptions %}
-{% from "views/layouts/components/header.njk" import header %}
 {% from "search-bar/search-bar.njk" import searchBar %}
 {% from "tables/tables.njk" import baseCaseListTable, baseManageHeaders, baseManageColumns %}
 
 {% block pageTitle %}
 	Case list - All Cases - Manage a Section 62A Development Application
-{% endblock %}
-
-{% block header %}
-	{{ header('Manage a Section 62A Application', baseUrl ) }}
 {% endblock %}
 
 {% block pageContent %}
--- a/apps/manage/src/util/config-middleware.js
+++ b/apps/manage/src/util/config-middleware.js
@@ -5,9 +5,51 @@
  */
 export function addLocalsConfiguration({ appName }) {
 	return (req, res, next) => {
+		const path = req.path;
+
+		const s62aLinks = [
+			{
+				text: 'Home',
+				href: '/s62a/cases'
+			},
+			{
+				text: 'Create new case',
+				href: '/s62a/cases/create-a-case/pre-application-or-application'
+			},
+			{
+				text: 'Sign out',
+				href: '/s62a/auth/signout'
+			}
+		];
+
+		const crownLinks = [
+			{
+				text: 'All Cases',
+				href: '/crown/cases'
+			},
+			{
+				text: 'Sign out',
+				href: '/auth/signout'
+			}
+		];
+
+		const links = res.locals.isS62A ? s62aLinks : crownLinks;
+
+		const headerTitle = res.locals.isS62A
+			? 'Manage Section 62A applications'
+			: 'Manage a Crown Development Application';
+
 		res.locals.config = {
-			appName
+			appName,
+			headerTitle: headerTitle,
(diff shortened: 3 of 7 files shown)


pinssbrett (changes requested):
I haven't done an in-depth review because I have a suggestion of a change in direction. Instead of building your own custom component, I think you should be able to use the generic header component from the lib `{% from "header/pins-header.njk" import pinsHeader %}` with `{ isExternal: false }`. If there are any overrides needed (i.e. does not work out of the box) please flag with me. That should save you building custom styles and templates and use our organisation-wide standard header

apps/manage/src/app/views/errors/401.njk
```diff
@@ -1,4 +1,13 @@
 {% extends "views/layouts/main.njk" %}
+{% from "header/pins-header.njk" import pinsHeader %}
+
+{% block header %}
+	{% if isS62A %}	
+		{{ pinsHeader({ isExternal: false }, config) }}
+	{% else %}
+		{{ super() }}
+	{% endif %}
+{% endblock %}
```
pinssbrett: instead of overriding the `header` block on all pages, can you handle it within `main.njk`? Using the same `isS62A` logic.

apps/manage/src/app/views/s62a/cases/list/controller.test.ts
```diff
@@ -7,6 +7,8 @@ import type { BaseLogger } from 'pino';
 import type { Request, Response } from 'express';
 import { configureNunjucks } from '../../../../nunjucks.js';
 
+const nunjucksEnv = configureNunjucks();
+
```
pinssbrett: Is this used?
```suggestion
```

apps/manage/src/app/views/layouts/components/core/banner.njk
pinssbrett, line 1: Is this required for the back office?

apps/manage/src/app/router.js
```diff
@@ -32,6 +32,7 @@ export function buildRouter(service) {
 	if (!service.authDisabled) {
 		service.logger.info('registering auth routes');
 		router.use('/auth', authRoutes);
+		router.use('/s62a/auth', authRoutes);
```
pinssbrett: Why is this needed? Wasn't auth working for S62A already?

apps/manage/src/util/config-middleware.js
```diff
@@ -5,9 +5,34 @@
...
 		res.locals.config = {
-			appName
+			appName,
+			headerTitle: 'Manage Section 62A applications',
+			isLive: true,
+			inBeta: false,
+			primaryNavigationLinks: links.map((link) => ({
+				...link,
+				current: link.href === path
+			}))
```
pinssbrett, line 50: This is adding some fixed values to both Crown and S62A - you should probably use an `isS62A` check to set different values here. Any values that are the same across both app types could be passed in from config similar to `appName`.

pinssbrett (changes requested):
Nice, and minimal! I have a few further suggestions and questions

apps/manage/src/app/router.js
andrewlilley-solirius: That's clearly slipped through from testing the initial custom-component version.  Apologies, will remove

apps/manage/src/app/views/errors/401.njk
andrewlilley-solirius: That's much cleaner, thank you!

apps/manage/src/app/views/s62a/cases/list/view.njk
```diff
@@ -2,16 +2,16 @@
...
+{% from "header/pins-header.njk" import pinsHeader %}
 
 {% block pageTitle %}
 	Case list - All Cases - Manage a Section 62A Development Application
 {% endblock %}
 
 {% block header %}
-	{{ header('Manage a Section 62A Application', baseUrl ) }}
+    {{ pinsHeader({ isExternal: false }, config) }}
 {% endblock %}
```
pinssbrett: You should now be able to remove the entire `header` block here as no need to override it any more

apps/manage/src/app/views/s62a/cases/view/view.njk
```diff
@@ -1,13 +1,18 @@
...
 {% set title = reference + " - " + journeyTitle + " - Manage a S62A Case" %}
 
 {% block pageTitle %}
 	{{ title }}
 {% endblock %}
 
+{% block header %}
+	{{ s62aHeader('Manage Section 62A applications', baseUrl, "/s62a/cases/create-a-case/pre-application-or-application" ) }}
+{% endblock %}
+
```
pinssbrett: Was this mistakenly included? Shouldn't need to adjust header here

apps/manage/src/util/config-middleware.js
```diff
@@ -5,9 +5,53 @@
...
+			}
+		];
+
+		const links = res.locals.isS62A ? s62aLinks : crownLinks;
+
+		const headerTitle = res.locals.isS62A
+			? 'Manage Section 62A applications'
+			: 'Manage a Crown Development Application';
+
+		console.log(res.locals.isS62A);
```
pinssbrett: Just a tidy up
```suggestion
```

pinssbrett (changes requested):
Just a couple more tweaks!

pinssbrett (approved):
Nice work!

PR #1091, opened 2026-07-14
chore(format): add EditorConfig and update formatting settings for consistency

AlexMoulin:
## Describe your changes
Changes were prompted by failing style-linting due to prettier updates.
This pull request primarily introduces and enforces consistent code style and line ending conventions across the project. It also simplifies several TypeScript union type definitions to improve readability.

**Code style and formatting improvements:**

* Added a new `.editorconfig` file to define consistent editor settings, UTF-8 encoding, LF line endings, and trimming trailing whitespace. Markdown files are configured to retain trailing whitespace. (`.editorconfig`)
* Updated `.prettierrc` to explicitly enforce LF (`endOfLine: "lf"`) line endings in Prettier formatting.
* Modified `.gitattributes` to ensure all text files use LF line endings by default (`* text=auto eol=lf`).

**Update dev dependencies**
Changes are a result of line formatting for prettier. Dev dependencies included for format-check to pass.

PR #1109, opened 2026-07-21
chore(project): format all nunjucks files with new editorconfig settings

pinssbrett:
## Describe your changes
Following from https://github.com/Planning-Inspectorate/crown-developments/pull/1091, we need to format all nunjucks files that have previously not been formatted now that `.editorconfig` has been added. Otherwise changes will be introduced on each commit that changes these.

The only changes in this PR should be whitespace and line breaks.

N/A


apps/manage/src/app/views/cases/list/view.njk
```diff
@@ -6,119 +6,121 @@
...
+			attributes: {
+			"aria-sort": "none"
+		}
+		},
+		{
+			text: 'Status',
+			attributes: {
+			"aria-sort": "none"
+		}
+		}
```
AlexMoulin: It looks like it still struggles with some objects

AlexMoulin (dismissed):
Looks good, there are some places where it formats the closing brackets weirdly but I think that can be fixed manually at some point when it comes up

AlexMoulin (approved):

PR #1111, opened 2026-07-21
Feat/peas 207 fees tab

oscargay:
## Describe your changes
- Adds the tab for `Fee`

https://pins-ds.atlassian.net/browse/PEAS-207

## Implementation
Screenshot 2026-07-21 at 13 05 59
Screenshot 2026-07-21 at 13 06 27
Screenshot 2026-07-21 at 13 06 35
Screenshot 2026-07-21 at 13 08 26
Screenshot 2026-07-21 at 13 08 36
Screenshot 2026-07-21 at 13 08 49
Screenshot 2026-07-21 at 13 08 57
Screenshot 2026-07-21 at 13 09 39

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
@@ -308,7 +308,7 @@ export class S62aCaseUpdateMapper {
 		return FEE_DATE_SET.has(key);
 	}
 
-	private isFeeStringField(key: string): key is (typeof FEE_DATE_FIELDS)[number] {
+	private isFeeStringField(key: string): key is (typeof FEE_STRING_FIELDS)[number] {
 		return FEE_STRING_SET.has(key);
 	}
 
--- a/apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
@@ -3,10 +3,21 @@ import { SITE_AREA_UNIT_ID } from '@pins/crowndev-database/src/seed/s62a/data-st
 import { viewModelToAddressUpdateInput } from '@pins/crowndev-lib/util/address.ts';
 import type { YesNo } from '@pins/crowndev-lib/util/types.ts';
 import { type Address, yesNoToBoolean } from '@planning-inspectorate/dynamic-forms';
-import { S62A_DATE_FIELDS } from './view-model.ts';
+import {
+	S62A_DATE_FIELDS,
+	FEE_BOOLEAN_FIELDS,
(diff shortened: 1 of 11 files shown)


apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
```diff
@@ -157,8 +223,95 @@ export class S62aCaseUpdateMapper {
...
+
+	private isFeeNumberField(key: string): key is (typeof FEE_NUMBER_FIELDS)[number] {
+		return FEE_NUMBER_SET.has(key);
+	}
+
+	private isFeeDateField(key: string): key is (typeof FEE_DATE_FIELDS)[number] {
+		return FEE_DATE_SET.has(key);
+	}
+
+	private isFeeStringField(key: string): key is (typeof FEE_DATE_FIELDS)[number] {
```
paymonsattar: I think this should be `typeof FEE_STRING_FIELDS`

oscargay: Good spot, I was on auto pilot here clearly... Will update and rebase onto main so this PR is clean!

paymonsattar (changes requested):
Just two small comments. Let me know what you think

apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
```diff
@@ -157,8 +223,95 @@ export class S62aCaseUpdateMapper {
...
+
+	/**
+	 * Creates the data on the Fees reference table
 	 */
+	private mapFees(input: Prisma.S62aCaseUpdateInput): void {
+		const feesToUpdate: Prisma.S62aFeesUpdateWithoutS62aCaseInput & Prisma.S62aFeesCreateWithoutS62aCaseInput = {};
+		let hasFeeUpdates = false;
+
+		for (const [key, value] of Object.entries(this.answers)) {
+			if (this.isFeeBooleanField(key)) {
```
paymonsattar, line 270: yesNoToBoolean(null) returns false, so clearing/removing a fee boolean via the remove-and-save path writes false to the DB instead of null - so this needs a nullish guard before the call

```
if (this.isFeeBooleanField(key)) {
	feesToUpdate[key] =
		value === null || value === undefined || value === ''
			? null
			: yesNoToBoolean(value);
	hasFeeUpdates = true;
}
```

oscargay: Neither of the fee boolean fields have the remove and save path currently, so I think will be something we add as and when we need to? What do you think?

paymonsattar (dismissed):

paymonsattar (approved):
###

PR #1115, opened 2026-07-22
Feat: s62a case details tab

paymonsattar:
## Describe your changes

- Adds details tab

https://pins-ds.atlassian.net/browse/PEAS-202

image
image

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/view/journey.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/journey.ts
@@ -25,6 +25,18 @@ export function createJourney(questions: Record<string, Question>, response: Jou
 	const isPlanningOrLbcCase = (response: JourneyResponse) =>
 		questionHasAnswer(response, questions.applicationType, APPLICATION_TYPE_ID.PLANNING_AND_LISTED_BUILDING_CONSENT);
 
+	const isApplicationCase = (response: JourneyResponse) =>
+		questionHasAnswer(response, questions.applicationPhase, PRE_APPLICATION_OR_APPLICATION_ID.APPLICATION);
+
+	const isPreApplicationCase = (response: JourneyResponse) =>
+		questionHasAnswer(response, questions.applicationPhase, PRE_APPLICATION_OR_APPLICATION_ID.PRE_APPLICATION);
+
+	const isApplicationLbcCase = (response: JourneyResponse) =>
+		isApplicationCase(response) && isPlanningOrLbcCase(response);
+
+	const isApplicationCilLiableCase = (response: JourneyResponse) =>
+		isApplicationCase(response) && questionHasAnswer(response, questions.cilLiable, BOOLEAN_OPTIONS.YES);
+
 	return new Journey({
 		journeyId: JOURNEY_ID,
 		sections: [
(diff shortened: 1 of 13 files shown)


apps/manage/src/app/views/s62a/cases/view/journey.ts
```diff
@@ -25,6 +25,15 @@ export function createJourney(questions: Record<string, Question>, response: Jou
 	const isPlanningOrLbcCase = (response: JourneyResponse) =>
 		questionHasAnswer(response, questions.applicationType, APPLICATION_TYPE_ID.PLANNING_AND_LISTED_BUILDING_CONSENT);
 
+	const isApplicationCase = (response: JourneyResponse) =>
+		questionHasAnswer(response, questions.applicationPhase, PRE_APPLICATION_OR_APPLICATION_ID.APPLICATION);
```
oscargay, line 29: Nice little helper, could be good to reuse it for the other checks that check this from the work I did before? Think there are a few on this page that could use it!

apps/manage/src/app/views/s62a/cases/view/questions.ts
```diff
@@ -47,6 +54,22 @@ export function getQuestions(answers: S62aCaseViewModel) {
 		? answers.applicationPhaseId
 		: PRE_APPLICATION_OR_APPLICATION_ID.APPLICATION;
 
+	const isPreApp = preAppOrAppPath === PRE_APPLICATION_OR_APPLICATION_ID.PRE_APPLICATION;
+	const statusOptions = isPreApp ? S62A_PRE_APPLICATION_STATUSES : S62A_APPLICATION_STATUSES;
+
+	// S62A only permits Hearing and Written representations
+	const s62aProcedures = APPLICATION_PROCEDURE.filter((p) => p.id !== APPLICATION_PROCEDURE_ID.INQUIRY);
```
oscargay: Just thinking out loud, is it a good idea to tie these 2 things together here? Probably fine if we are confident Crown's procedures are never going to change. Either way probably a good idea to do the opposite check here, so we don't have to remember to update this line if Crown ever decides to add a future item to its procedure list?

```
const s62aProcedures = APPLICATION_PROCEDURE.filter((p) => (p.id === APPLICATION_PROCEDURE_ID.HEARING) || (p.id === APPLICATION_PROCEDURE_ID.WRITTEN_REPS));
```

What do you think?

apps/manage/src/app/views/s62a/cases/view/view-model.test.ts
```diff
@@ -66,7 +66,18 @@ describe('s62aCaseToViewModel', () => {
 				start: startDate,
 				end: endDate
 			},
-			representationsPublishDate: publishDate
+			representationsPublishDate: publishDate,
```
oscargay: I think actually looking at this now we should change this test and the next one, otherwise we are going to be constantly adding keys for no reason like you have had to do here. Maybe it's better just to do a sample of relevant keys. Maybe this test should just test like reference, description, siteIsVisibleFromPublicLand, etc.

And likewise for the next one only the next one is checking nulls -> undefined so just need to check that?

What are your thoguhts?

packages/database/src/seed/s62a/data-static.ts
```diff
@@ -62,22 +62,74 @@ export const SITE_AREA_UNITS = [
...
 ];
 
-/**
- * NB. statuses are separate to crown as they are
- * going to be different.
- *
- * TODO: once full list of statuses confirmed,
- * update the list.
- */
 export const S62A_STATUS_ID = Object.freeze({
```
oscargay, line 65: NIT: Just wondering why you split out the keys like this with the comments, it's just an object so they're unordered, the two arrays after help to say which ones are for app or pre-app well enough I think! Unless I'm missing an intention here?

packages/database/src/seed/data-static.ts
```diff
@@ -531,101 +533,109 @@ export const CATEGORIES: Prisma.CategoryCreateInput[] = [
 
 type UpsertReferenceDataArgs =
 	| {
-			delegate: Prisma.ApplicationDecisionOutcomeDelegate;
-			input: Prisma.ApplicationDecisionOutcomeCreateInput;
-	  }
+		delegate: Prisma.ApplicationDecisionOutcomeDelegate;
```
oscargay: I'm guessing these things might be the formatting issue? It always tries to change it for me whenever I save?

oscargay (changes requested):
Looks good! A few comments but they're all quite small nitpicky things. Let me know what you think?

packages/lib/forms/custom-components/cil-amount/cil-amount-length-validator.js
```diff
@@ -0,0 +1,29 @@
+import BaseValidator from '@planning-inspectorate/dynamic-forms/src/validator/base-validator.js';
+import { body } from 'express-validator';
+import { BOOLEAN_OPTIONS } from '@planning-inspectorate/dynamic-forms/src/components/boolean/question.js';
+
+/**
+ * Validates that a CIL amount fits the Decimal(12,2) column:
+ * up to 10 digits before the decimal point and up to 2 after.
+ * Runs alongside CILAmountValidator; does not replace it.
+ */
+export default class CILAmountLengthValidator extends BaseValidator {
```
oscargay: Just realised we probably need a test for this, also it's in JS rather than TS?

apps/manage/src/app/views/s62a/cases/view/view-model.test.ts
paymonsattar: I think this is a good idea! I'll implement something now

packages/database/src/seed/data-static.ts
paymonsattar: yes, I believe so - I've found closing the file and then running the format works as you don't save on top. Could be worth investigating as it seems prettier is getting confused here

apps/manage/src/app/views/s62a/cases/view/questions.ts
paymonsattar: I think your suggestion is more safe, have updated it but I've used `include` rather than `filter` as I think it reads better

oscargay (approved):
LGTM! There's one comment left but seeing as it was a small nitpick, happy to approve :)

PR #1122, opened 2026-07-23
Feat/peas 359 adds contacts questions

oscargay:
## Describe your changes
- Adds base contact questions without other contacts and with no patch or post logic
- Breaking up into smaller PRs

## Issue ticket number and link
https://pins-ds.atlassian.net/browse/PEAS-359

PR #1123, opened 2026-07-23
Feat/peas 360 adds update and view logic

oscargay:
## Describe your changes
- Adds the logic for saving and retrieving contacts
- Branched from https://github.com/Planning-Inspectorate/crown-developments/pull/1122 which needs to be merged first
- I have tried to break the ticket down already, but this one is still quite large, as such I have made sure to commit in clear a direct commits that should make the review process simpler.

https://pins-ds.atlassian.net/browse/PEAS-360

## Implementation
Screenshot 2026-07-24 at 11 51 07
Screenshot 2026-07-24 at 11 52 19
Screenshot 2026-07-24 at 11 52 48

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/view/constants.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/constants.ts
@@ -19,3 +19,57 @@ export const CASE_DETAILS_QUESTION_TEXT = {
 		applicationReceivedDateQuestion: 'When was the application received?'
 	}
 };
+
+/**
+ * The consistent select query that is used when generating
+ * the view model for the case details page.
+ */
+export const S62A_VIEW_SELECT_INCLUDE = {
+	Lpa: {
+		include: {
+			Address: true
+		}
+	},
+	SecondaryLpa: {
+		include: {
+			Address: true
+		}
+	},
+	S62aStatus: true,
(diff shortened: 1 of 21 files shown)


apps/manage/src/app/views/s62a/cases/view/constants.ts
```diff
@@ -19,3 +19,57 @@ export const CASE_DETAILS_QUESTION_TEXT = {
 		applicationReceivedDateQuestion: 'When was the application received?'
 	}
 };
+
+/**
+ * The consistent select query that is used when generating
+ * the view model for the case details page.
+ */
+export const S62A_VIEW_SELECT_INCLUDE = {
```
paymonsattar, line 28: I think you should add `as const satisfies Prisma.S62aCaseInclude` here otherwise you get a looser payload type than the query actually returns. You didn't have this which is why you needed to add `Role: true' in the next PR [here](https://github.com/Planning-Inspectorate/crown-developments/pull/1127/changes#diff-139f59b4900c309c8448ed5d2cc477b69c7933a13e1b3f8eade7178f7de8ddaeR63)

paymonsattar (commented):
Small comment, what do you think?

paymonsattar (approved):

PR #1126, opened 2026-07-27
fix/peas 241 update lpa error messaging

andrewlilley-solirius:
## Describe your changes
Updated error messages that are presented in the S62A manage back office LPA Contact Details pages as described in the Jira ticket

Updated TelephoneNumberValidator class in telephone-number-validator.ts to accept custom error message strings and amended telephone-number-validator.test.ts 

https://pins-ds.atlassian.net/browse/PEAS-241


apps/manage/src/app/views/s62a/cases/create-a-case/question-factories.ts
```diff
@@ -56,9 +57,12 @@ export const createLpaContactQuestion = (isSecondary: boolean) => {
...
 						validators: [
-							new RequiredValidator(`Enter ${labelPrefix} contact's email address`),
+							new RequiredValidator(`Enter email address of the ${questionText} contact`),
 							new StringValidator({
-								maxLength: { maxLength: 250, maxLengthMessage: 'Email address must be between 3 and 250 characters' },
+								maxLength: {
+									maxLength: 250,
+									maxLengthMessage: `Email of the ${questionText} contact must be 250 characters or less`
+								},
 								minLength: { minLength: 3, minLengthMessage: 'Email address must be between 3 and 250 characters' }
```
oscargay: Do we still want to be keeping the `minLength` check if we are updating the messaging to be 250 or less, or can this be removed? To be honest, I don't know if the min check even does anything, because 3 characters would be an invalid email address anyway of course

packages/lib/validators/telephone-number-validator.js
```diff
@@ -11,10 +11,13 @@ export default class TelephoneNumberValidator extends StringValidator {
...
 	 * @param {string} [params.fieldName]
+	 * @param {Object} [params.maxLength]
+	 * @param {number} [params.maxLength.maxLength]
+	 * @param {string} [params.maxLength.maxLengthMessage]
 	 */
-	constructor({ fieldName } = {}) {
+	constructor({ maxLength, fieldName } = {}) {
 		super({
-			maxLength: { maxLength: 15, maxLengthMessage: 'Phone number must be 15 characters or less' },
+			maxLength: maxLength || { maxLength: 15, maxLengthMessage: 'Phone number must be 15 characters or less' },
```
oscargay: I think this would make more sense if the two keys were decoupled, e.g.

```
maxLength: { maxLength: maxLength.maxLength || 15, maxLengthMessage: maxLength.maxLengthMessage || 'Phone number must be 15 characters or less' },
```
Naming would have to be adjusted so it isn't so confusing! (maxLengthParams?)

Just because otherwise right now a person can pass in an incomplete object (or anything "truthy") and it would break, whereas this new one would allow someone to just pass a length if they wanted to change that, or pass in a new message, or both!

Seeing as this is quite a little class, perhaps it is time to turn it into TS, would make some of the stuff above a lot more manageable. What do you think?

oscargay (changes requested):
Nice work Andrew, a couple comments for your consideration! Good job

packages/lib/validators/telephone-number-validator.js
andrewlilley-solirius: That definitely makes sense, and it certainly wouldn't take much to port that over to TypeScript as well. Will do!

apps/manage/src/app/views/s62a/cases/create-a-case/question-factories.ts
andrewlilley-solirius: Definitely agree that it would be a lot easier to just remove it, given it’s otherwise redundant and doubt there's much performance gain by having it fail validation before running the email regex checks

oscargay (approved):
LGTM!

PR #1129, opened 2026-07-27
feat: add case team tab

paymonsattar:
## Describe your changes
This PR adds the case teams tab.
 - update s62a table and create migration
 - Add logic to controller, questions config and view model
 - Add some helper functions to help with correctly saving list items within manage list

https://pins-ds.atlassian.net/browse/PEAS-206

Screenshot 2026-07-27 at 18 44 02
Screenshot 2026-07-27 at 18 45 16
Screenshot 2026-07-27 at 18 45 21
Screenshot 2026-07-27 at 18 45 24
Screenshot 2026-07-27 at 18 45 29
Screenshot 2026-07-27 at 18 45 32
Screenshot 2026-07-27 at 18 46 05

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/view/journey.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/journey.ts
@@ -196,6 +196,18 @@ export function createJourney(questions: Record<string, Question>, response: Jou
 				)
 				.addQuestion(questions.representationsPeriod)
 				.addQuestion(questions.representationsPublishDate),
+			new Section('', 'case-team')
+				.withSectionCondition(() => currentTab === VIEW_TAB_ID.CASE_TEAM)
+				.addQuestion(
+					questions.manageCaseTeamInspectors,
+					new ManageListSection().addQuestion(questions.inspectorId).addQuestion(questions.inspectorAllocatedDate)
+				)
+				.addQuestion(questions.caseOfficer)
+				.addQuestion(questions.assessorInspector)
+				.startMultiQuestionCondition('is-application', isApplicationCase)
+				.addQuestion(questions.planningOfficer)
+				.addQuestion(questions.reader)
+				.endMultiQuestionCondition('is-application'),
 			new Section('', 'fee')
 				.withSectionCondition(() => currentTab === VIEW_TAB_ID.FEE)
 
--- a/apps/manage/src/app/views/s62a/cases/view/questions.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/questions.ts
@@ -41,7 +41,7 @@ import { SEPARATOR_TYPE } from '@pins/crowndev-lib/forms/custom-components/custo
 import MultiFieldInputValidator from '@pins/crowndev-lib/validators/multi-field-input-validator.js';
 import { CASE_DETAILS_QUESTION_TEXT } from './constants.ts';
 import { getApplicantContactsValidator, isApplicationType } from '../util/questions.ts';
-import { getLpaOptions } from '@pins/crowndev-lib/util/questions.ts';
+import { getLpaOptions, referenceDataToRadioOptions } from '@pins/crowndev-lib/util/questions.ts';
 import CustomDatePeriodValidator from '@pins/crowndev-lib/validators/custom-date-period-validator.js';
 import FeeAmountValidator from '@pins/crowndev-lib/forms/custom-components/fee-amount/fee-amount-validator.js';
 import CILAmountValidator from '@pins/crowndev-lib/forms/custom-components/cil-amount/cil-amount-validator.js';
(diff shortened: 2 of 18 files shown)


packages/database/src/schema.prisma
```diff
@@ -243,6 +243,25 @@ model S62aCase {
   cilAmount               Decimal? @db.Decimal(12, 2)
   /// whether exempt from biodiversity net gain (BNG)
   bngExempt               Boolean?
+
+  /// Case Team fields
+  /// Entra ID
+  inspector1Id            String?
```
oscargay: Just having a think about this, I think the data model that was done on the spreadsheet probably didn't take into account this being an "add to list" / followed what Crown did (but I think Crown did it that way because they didn't have a good way of doing add to list?). 

In my head this makes more sense as a 1-many join, then you can have however many you want and the date table will just go on that? And then you won't have to do any specific stuff to do with adding like `-1` `-2` etc. onto the view model to get it to work with the list? 

What do you think? Might be worth checking in with Abbas to see if there was a reason to do it in the 3 columns as well 🤔

packages/database/src/schema.prisma
```diff
@@ -243,6 +243,25 @@ model S62aCase {
...
+  /// Entra ID
+  inspector2Id            String?
+  inspector2AllocatedDate DateTime?
+  /// Entra ID
+  inspector3Id            String?
+  inspector3AllocatedDate DateTime?
+  /// Entra ID - Inspector assigned to help with a specialist issue
+  assessorInspectorId     String?
+  /// Entra ID
+  caseOfficerId           String?
```
oscargay: I remember there was some talk about making a `User` table similar to `MPESC` so that we are more versatile with adding columns in the future (e.g. a name column), I can't remember where we landed on it though?

oscargay (commented):
Just a couple thoughts on only the data model side whilst I hold off on the rest due to overlapping work! Would be keen to get your thoughts on these, just worried that whatever decision we take now it could bottleneck us in the future so want to make sure it's the right decision (even though what is here is what is specified in the spreadsheet!). Thanks

packages/database/src/schema.prisma
paymonsattar: Have checked with Sophie and Charlotte and we are ok to go ahead with the one to many join. I've now created a table for this `S62aCaseInspector`.

packages/database/src/schema.prisma
paymonsattar: Have discussed this with the team and decided to move forward with this idea. Thanks for spotting this!

apps/manage/src/app/views/s62a/cases/view/journey.ts
```diff
@@ -196,6 +196,18 @@ export function createJourney(questions: Record<string, Question>, response: Jou
...
+				.addQuestion(
+					questions.manageCaseTeamInspectors,
+					new ManageListSection().addQuestion(questions.inspectorId).addQuestion(questions.inspectorAllocatedDate)
+				)
+				.addQuestion(questions.caseOfficer)
+				.addQuestion(questions.assessorInspector)
+				.addQuestion(questions.planningOfficer)
+				.withCondition(isApplicationCase)
+				.addQuestion(questions.reader)
+				.withCondition(isApplicationCase),
```
oscargay: NIT: if there are >1 I think it's better practice to do a multi condition (even though only 2!)

apps/manage/src/app/views/s62a/cases/view/questions.ts
```diff
@@ -56,7 +56,16 @@ type ApplicantOrg = {
...
 
-export function getQuestions(answers: S62aCaseViewModel, isQuestionView?: boolean) {
+export function getQuestions(
+	answers: S62aCaseViewModel,
+	{
+		isQuestionView,
+		groupMembers
+	}: {
+		isQuestionView?: boolean;
+		groupMembers: { caseOfficers: unknown[]; inspectors: unknown[] };
```
oscargay: I think if I remember correctly there's an interface already in Crown that has these typed already so can use that `groupMembers: <real_type>`

apps/manage/src/app/views/s62a/cases/view/questions.ts
```diff
@@ -1239,6 +1248,86 @@ export function getQuestions(answers: S62aCaseViewModel, isQuestionView?: boolea
...
+		inspectorId: {
+			type: COMPONENT_TYPES.SELECT,
+			title: 'Inspector',
+			question: 'Which inspector is assigned to this case?',
+			fieldName: 'inspectorId',
+			url: 'inspector',
+			validators: [new RequiredValidator('Select an inspector')],
+			options: referenceDataToRadioOptions(groupMembers.inspectors as Reference[], true)
+		},
+		inspectorAllocatedDate: {
```
oscargay, line 1276: On other dates, I've been asked to put in a hint text, might be worth checking with the BA team if we want that here?

oscargay (commented):
Looks good! A couple small comments & questions but nothing major. Let me know what you think?

apps/manage/src/app/views/s62a/cases/view/questions.ts
paymonsattar: Added a hint!

apps/manage/src/app/views/s62a/cases/view/questions.ts
paymonsattar: Good shout, I've added this now

apps/manage/src/app/views/s62a/cases/view/journey.ts
paymonsattar: Good shout, missed this

oscargay (approved):
LGTM!

PR #1135, opened 2026-07-28
feat: add eia tab

paymonsattar:
## Describe your changes
Adds EIA tab to view & edit case

https://pins-ds.atlassian.net/browse/PEAS-208

Screenshot 2026-07-28 at 14 39 55
Screenshot 2026-07-28 at 14 40 22
Screenshot 2026-07-28 at 14 40 30
Screenshot 2026-07-28 at 14 40 35
Screenshot 2026-07-28 at 14 40 55

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
@@ -128,6 +128,11 @@ export interface UpdateCaseAnswers {
 	manageApplicantContactDetails?: ApplicantContactAnswer[];
 
 	manageAdditionalContacts?: AdditionalContactAnswer[];
+
+	// EIA tab
+	eiaScreening?: boolean | YesNo | null;
+	eiaScreeningOutcome?: boolean | YesNo | null;
+	environmentalStatementReceivedDate?: Date | null;
 }
 
 /**
@@ -159,6 +164,7 @@ export class S62aCaseUpdateMapper {
 		this.mapFees(input);
 		this.mapLpaContacts(input);
 		this.mapApplicantsAndAgents(input);
+		this.mapEiaScalars(input);
 
 		return input;
 	}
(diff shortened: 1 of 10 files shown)


apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
```diff
@@ -362,6 +367,22 @@ export class S62aCaseUpdateMapper {
...
 
+	/**
+	 * Maps the EIA tab scalar fields.
+	 * environmentalStatementReceivedDate is handled by mapDates via S62A_DATE_FIELDS.
+	 */
+	private mapEiaScalars(input: Prisma.S62aCaseUpdateInput): void {
+		const ans = this.answers;
+
+		if (this.hasAnswer('eiaScreening')) {
+			input.eiaScreening = ans.eiaScreening == null ? null : yesNoToBoolean(ans.eiaScreening);
```
oscargay: if `eiaScreening` is coming through like other boolean questions (as actual boolean values) than could we not do a similar check to those ones? `typeof <x> === 'boolean'`

oscargay (commented):
Looks good, just 1 question but otherwise all looks good!

Will need to rebase & possibly re-review because we've had a couple merges in this area

apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
pinssbrett: It may be that types in `UpdateCaseAnswers` need tightening up - any `YesNo` values are likely true `booleans` instead due to how dynamic-forms handles boolean questions (but should remain as `YesNo` in the view model)

paymonsattar: Both correct! `COMPONENT_TYPES.BOOLEAN` submits real booleans, so I've switched to the `typeof === 'boolean'` guard to match the other boolean fields, and typed the `EIA` fields as `boolean | null` in `UpdateCaseAnswers` while keeping `YesNo` in the view model.

In regards to your comment @pinssbrett - I agree with you. I'll make a note to do a clean up after these tickets are merged in a seperate PR. I will update this comment with the PR link once that PR is up.

oscargay (approved):
LGTM!

PR #1136, opened 2026-07-28
feat: adds folders to s62a and landing page

oscargay:
## Describe your changes
- adds the folder generation and landing page

https://pins-ds.atlassian.net/browse/PEAS-59

## Implementation
Screenshot 2026-07-28 at 16 19 35
Screenshot 2026-07-28 at 16 19 37

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/view/controller.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/controller.ts
@@ -39,7 +39,8 @@ export function buildViewCaseDetails(): AsyncRequestHandler {
 			viewTabs: viewTabsToShow,
 			// URL without the /:tab slug, needed for routing uses in FE.
 			cleanUrl: `/s62a/cases/${id}`,
-			banner
+			banner,
+			foldersUrl: `/s62a/cases/${id}/case-folders`
 		});
 	};
 }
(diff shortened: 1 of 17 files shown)


apps/manage/src/app/views/s62a/cases/view/controller.ts
```diff
@@ -146,3 +148,15 @@ export function clearCaseUpdatedSession(req: Request, id: string): void {
 function getJourneyAnswers(res: Response): S62aCaseViewModel | undefined {
 	return res.locals.journeyResponse?.answers as unknown as S62aCaseViewModel;
 }
+
+/**
+ * Validate the format of the id parameter
+ */
+export function validateIdFormat(req: Request, res: Response, next: NextFunction) {
+	const id = getStringParam(req.params, 'id');
```
pinssbrett, line 156: I think this will throw before ever getting to `notFoundHandler` - should it be optional in order to handle more gracefully?
```suggestion
	const id = getOptionalStringParam(req.params, 'id');
```

pinssbrett (changes requested):
Looks good, I think  just one small thing to address

apps/manage/src/app/views/s62a/cases/view/controller.ts
oscargay: Thanks! It's a good point, I wasn't sure because `getStringParam` seems to be the typical pattern I've seen used for checking for the ID, as it's not really optional like some fields could be.

I think the discrepancy is the two scenarios are:
(1) id is not passed at all or is sent in a completely wrong type = error and explain it needs to be a string
(2) id is sent in the correct format but it's not in the desired UUID format, so would cause an ugly prisma error because of this, so we do a nice 404, as that's what would have happened if it could query the DB.

So I think it does make some sense, what do you think?

pinssbrett: ah okay that granularity makes sense. agreed!

pinssbrett (approved):

PR #1149, opened 2026-07-31
feat(manage): add long string field support for history

RobBroadhead30:
## Describe your changes
This pull request introduces support for improved audit logging and display for long text fields (such as "description" and "costsApplicationsComment") in the case history view. It ensures that updates to these fields are recorded with a dedicated action and rendered in the UI with expandable old/new value sections for better readability. The changes also include thorough testing and style updates.

**Audit logging and field tracking:**

* Added a new audit action `FIELD_UPDATED_LONG` and corresponding template for long text fields, ensuring that updates to fields like `description` and `costsApplicationsComment` are tracked distinctly from standard fields. [[1]](diffhunk://#diff-3f1a0f1a8311f5cf5608c8076c11937480e21578a7f91f75f31419503bf2752bR20) [[2]](diffhunk://#diff-3f1a0f1a8311f5cf5608c8076c11937480e21578a7f91f75f31419503bf2752bR54) [[3]](diffhunk://#diff-6e26e0918f33a6815dfe31b2f9f6330d01a493c8d7442ae5515860b08f50f381R75) [[4]](diffhunk://#diff-6e26e0918f33a6815dfe31b2f9f6330d01a493c8d7442ae5515860b08f50f381L109-R119) [[5]](diffhunk://#diff-6e26e0918f33a6815dfe31b2f9f6330d01a493c8d7442ae5515860b08f50f381L493-R515)
* Updated the field resolver to treat `description` and `costsApplicationsComment` as long text fields.

**Case history view and UI improvements:**

* Enhanced the case history view model and template to include a `longField` property for long text changes, rendering old and new values in expandable details sections using the GOV.UK details component. [[1]](diffhunk://#diff-353bf590bc6bef93507b91b03ad41ddd3c057062e5fe5658f32d4f4205163707L18-R41) [[2]](diffhunk://#diff-f0765011ae2a99a64c93313303eb5e555b89cc94e7294c4128f74b66975aa44dR40-R54) [[3]](diffhunk://#diff-f0765011ae2a99a64c93313303eb5e555b89cc94e7294c4128f74b66975aa44dR80) [[4]](diffhunk://#diff-f0765011ae2a99a64c93313303eb5e555b89cc94e7294c4128f74b66975aa44dR103) [[5]](diffhunk://#diff-924df64c93df563f043b122ec0d53cbb96002a3a6c828b93d5d48123f7c06480R1-R25) [[6]](diffhunk://#diff-ff5d1763b61f9fd7512cd80408646563d4482a20e56251eee70155c52bf0792cR17)

**Testing:**

* Added comprehensive tests for the new audit action, the view model logic, and the audit recording process to ensure correct handling and display of long text field updates. [[1]](diffhunk://#diff-692901b63f191eddd9155b3a049e062d3e70c2703d7d5e91b267a7d2cca1b6a3L3-R3) [[2]](diffhunk://#diff-692901b63f191eddd9155b3a049e062d3e70c2703d7d5e91b267a7d2cca1b6a3R76-R89) [[3]](diffhunk://#diff-692901b63f191eddd9155b3a049e062d3e70c2703d7d5e91b267a7d2cca1b6a3R104-R113) [[4]](diffhunk://#diff-a8a0088bd0c6de44c57e6b11de7c6e3121c4d2b87ef69c750375c1496110e8c6R124-R218) [[5]](diffhunk://#diff-ea897f553431c94f23910d1d49e5f869434290cbe25eb5b518d86c7d243f7ad3R2847-R2988)

These changes collectively improve the traceability and user experience for editing long text fields in cases.
https://pins-ds.atlassian.net/browse/CROWN-1634

image

image

image


apps/manage/src/app/views/cases/case-history/components/case-history-table.scss
```diff
@@ -0,0 +1,25 @@
+.pins-case-history-table {
+	table-layout: fixed;
+	width: 100%;
+
+	.govuk-table__header,
+	.govuk-table__cell {
+		vertical-align: top;
+		overflow-wrap: anywhere;
+	}
```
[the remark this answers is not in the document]
RobBroadhead30: This suggestion adds too much of a gap for each section in the case history

apps/manage/src/app/views/cases/case-history/view.njk
```diff
@@ -36,30 +37,46 @@
...
-										{% for file in row.files %}
-											<li>{{ file }}</li>
-										{% endfor %}
-									</ul>
-								{% endset %}
+							{% if row.longField %}
+								{{ row.details }}
+								{% if row.longField.oldValue %}
+									{{ govukDetails({
+										summaryText: "previous application answer",
```
pinssbrett: This needs to say `Previous [fieldName]` as per the acceptance criteria

```suggestion
										summaryText: ["Previous ", row.longField.fieldName] | join,
```

apps/manage/src/app/views/cases/case-history/view.njk
```diff
@@ -36,30 +37,46 @@
...
+										html: row.longField.oldValue | escape | replace("\n", "<br>") | safe
+									}) }}
+								{% endif %}
+								{% if row.longField.newValue %}
+									{{ govukDetails({
+										summaryText: "new application answer",
```
pinssbrett: Ditto
```suggestion
										summaryText: ["New ", row.longField.fieldName] | join,
```

apps/manage/src/app/audit/resolvers/field-resolver.ts
```diff
@@ -135,6 +135,21 @@ function booleanResolver(fieldName: string): FieldResolver {
...
+function longStringResolver(fieldName: string): FieldResolver {
+	return {
+		resolve(previousCase, newAnswer) {
+			return {
+				oldValue: formatValue(previousCase[fieldName]),
+				newValue: formatValue(newAnswer)
+			};
+		}
+	};
+}
```
pinssbrett: Is there any difference between this and the defaultResolver? Is it needed?

apps/manage/src/app/views/cases/case-history/view-model.ts
```diff
@@ -13,19 +25,32 @@ export interface CaseHistoryRow {
...
 }
 
 /**
  * Transforms raw audit events into rows ready for the case history table.
  */
 export function createCaseHistoryViewModel(events: Array<AuditEvent & { userName: string }>): CaseHistoryRow[] {
 	return events.map((event) => {
+		const { action, metadata, createdAt, userName } = event;
+		const storedFieldName = metadata?.fieldName as string | undefined;
+		const isLongField = storedFieldName && LONG_FIELD_DISPLAY_NAMES.has(storedFieldName);
```
pinssbrett: You're checking against field name in two places - here and in update-case. These two definitions could drift. Instead, here you already have the action, so could you check for `AUDIT_ACTIONS.FIELD_UPDATED_LONG`?

apps/manage/src/app/views/cases/case-history/view-model.ts
```diff
@@ -13,19 +25,32 @@ export interface CaseHistoryRow {
...
+				: isAuditAction(action)
+					? resolveTemplate(action, metadata ?? undefined)
+					: `Unknown action: ${action}`,
+			user: userName,
+			...(isLongField && {
+				longField: {
+					oldValue: (metadata?.oldValue as string) || '-',
+					newValue: (metadata?.newValue as string) || ''
+				}
+			})
```
pinssbrett, line 41: The `longField` handling in the view model also breaks the clean abstraction where `resolveTemplate` was the single place for all action-specific logic. Consider extending `resolveTemplate` to return a structured object instead of just a string, moving the `FIELD_UPDATED_LONG` special case into `resolveTemplate` itself. This would let the view model remain a generic transformation layer that simply passes through whatever `resolveTemplate` returns, making it easier to add future action types that need extra data.

apps/manage/src/app/audit/actions.ts
```diff
@@ -17,6 +17,7 @@ export const AUDIT_ACTIONS = {
 	FIELD_SET: 'FIELD_SET',
 	FIELD_UPDATED: 'FIELD_UPDATED',
 	FIELD_CLEARED: 'FIELD_CLEARED',
+	FIELD_UPDATED_LONG: 'FIELD_UPDATED_LONG',
```
pinssbrett, line 20: What about when the field is set (i.e. no previous value)? And maybe cleared, but I don't think any long fields have that functionality

apps/manage/src/app/views/cases/view/update-case.ts
```diff
@@ -496,6 +496,11 @@ async function recordAuditEntries(
 				action = AUDIT_ACTIONS.FIELD_UPDATED;
 			}
 
+			const LONG_AUDIT_FIELDS = new Set(['description', 'healthAndSafetyIssue']);
```
pinssbrett: I think clearer to define this near `AUDITABLE_SCALAR_FIELDS` instead of within the loop

pinssbrett: Health and safety issue is not a long field and not in the acceptance criteria for this ticket. Costs applications comment is - did you mean that?

apps/manage/src/app/views/cases/case-history/components/case-history-table.scss
```diff
@@ -0,0 +1,25 @@
...
+	.govuk-table__header:nth-child(2),
+	.govuk-table__cell:nth-child(2) {
+		width: 58%;
+	}
+
+	.govuk-table__header:nth-child(3),
+	.govuk-table__cell:nth-child(3) {
+		width: 20%;
+	}
+}
```
pinssbrett, line 25: This uses magic numbers to fix the table ratios, instead of allowing the browser to fit content automatically. What was wrong without it?

pinssbrett (changes requested):
Good work, but some things need addressing

apps/manage/src/app/views/cases/case-history/components/case-history-table.scss
RobBroadhead30: The issue was extending the show and hide drop downs as it was shifting over to the left when extended.. with my previous fix the table still looked odd so each column now has a set width no matter the screen size

pinssbrett: ah right, good catch! this is the best solution then, but I'd advise adding a comment to explain why it's needed.
also move this file to `sass/components/_pins-case-history-table.scss` so it's with other components (once rebased to new style structure in `main`)

apps/manage/src/app/audit/actions.ts
pinssbrett: I don't think this has been addressed - there is still no audit for a field being added. I've double-checked and `costsApplicationsComment` can be both added and deleted. I'd also bring the naming convention in line with other field types like on MPESC. Therefore:
- `LONG_FIELD_SET`
- `LONG_FIELD_UPDATED`
- `LONG_FIELD_CLEARED`

Display only `newValue` for SET and `oldValue` for CLEARED

apps/manage/src/app/views/cases/case-history/view-model.ts
```diff
@@ -15,17 +15,30 @@ export interface CaseHistoryRow {
...
 	return events.map((event) => {
+		const { action, metadata, createdAt, userName } = event;
+		const dateTimeFormatted = formatDateTime(new Date(createdAt));
+
+		if (!isAuditAction(action)) {
+			return { dateTimeFormatted, details: `Unknown action: ${action}`, user: userName };
+		}
+
+		const details = resolveTemplate(action, metadata ?? undefined);
+		const isLong = metadata?.isLong === true;
```
pinssbrett, line 29: you can tell this from `action`, so using that to check instead of passing `isLong` saves you from having two strands of logic again

apps/manage/src/app/views/cases/view/update-case.ts
```diff
@@ -490,25 +493,26 @@ async function recordAuditEntries(
...
-			if (oldValue === '-') {
+			if (isLongField) {
+				action = AUDIT_ACTIONS.FIELD_UPDATED_LONG;
+			} else if (oldValue === '-') {
 				action = AUDIT_ACTIONS.FIELD_SET;
 			} else if (newValue === '-') {
 				action = AUDIT_ACTIONS.FIELD_CLEARED;
 			} else {
 				action = AUDIT_ACTIONS.FIELD_UPDATED;
 			}
```
pinssbrett, line 506: This is where we may have to diverge from MPESC a bit, because MPESC only has one set of scalar field actions and we'll have more than that. You'll need to calculate the difference between SET, CLEARED and UPDATED for long fields too. Because this will end up calculating `action` from both `fieldName`, `oldValue` and `newValue`, I recommend refactoring this to a method that returns `action` based on those values. that will save you from a big chain of `if` statements

pinssbrett (changes requested):
Great work! I have a couple of notes on my previous comment

apps/manage/src/app/views/cases/case-history/view.njk
```diff
@@ -36,30 +37,46 @@
...
+										summaryText: ["Previous ", row.longField.fieldName] | join,
+										html: row.longField.oldValue | escape | replace("\n", "<br>") | safe
+									}) }}
+								{% endif %}
+								{% if row.longField.newValue %}
+									{{ govukDetails({
+										summaryText: ["New ", row.longField.fieldName] | join,
+										html: row.longField.newValue | escape | replace("\n", "<br>") | safe
+									}) }}
+								{% endif %}
```
pinssbrett, line 53: This needs spacing between it and `row.details` - use a wrapper?

pinssbrett (changes requested):

apps/manage/src/app/sass/components/_pins-case-history-table.scss
```diff
@@ -0,0 +1,25 @@
...
+	.govuk-table__header:nth-child(2),
+	.govuk-table__cell:nth-child(2) {
+		width: 58%;
+	}
+
+	.govuk-table__header:nth-child(3),
+	.govuk-table__cell:nth-child(3) {
+		width: 20%;
+	}
+}
```
pinssbrett: still needs a comment

apps/manage/src/app/views/cases/view/update-case.ts
```diff
@@ -508,7 +504,8 @@ async function recordAuditEntries(
 				metadata: {
 					fieldName: getFieldDisplayName(fieldName),
 					oldValue,
-					newValue
+					newValue,
+					...(action === AUDIT_ACTIONS.FIELD_UPDATED_LONG ? { isLong: true } : {})
```
pinssbrett: You still have the `isLong` being passed as metadata when you are now checking the action - this can be removed (from tests as well)

apps/manage/src/app/views/cases/case-history/view.njk
```diff
@@ -36,30 +37,53 @@
...
-							{% if row.files and row.files | length > 0 %}
-								{% set filesListHtml %}
-									<ul class="govuk-list govuk-list--bullet">
-										{% for file in row.files %}
-											<li>{{ file }}</li>
-										{% endfor %}
-									</ul>
-								{% endset %}
+							{% if row.longField %}
+								<div class="pins-case-history-long-field">
```
pinssbrett: Is this class and `pins-case-history-long-field__details` being used for anything?

apps/manage/src/app/audit/actions.ts
pinssbrett: Because you're using the default field actions for SET and CLEARED, the content is still being displayed inline:
image
Instead you need to add all three actions for long fields, so that the field content always uses the accordions to display. I'd also bring the naming convention in line with other field types like on MPESC so that the action itself is always at the end of the name. Therefore:
- `LONG_FIELD_SET`
- `LONG_FIELD_UPDATED`
- `LONG_FIELD_CLEARED`

Display only `newValue` for SET and `oldValue` for CLEARED

apps/manage/src/app/views/cases/view/update-case.ts
pinssbrett: This has been partially addressed but your handling will need to cover all three long field action types (see other comment)

apps/manage/src/app/audit/actions.ts
```diff
@@ -51,6 +56,11 @@ export const AUDIT_TEMPLATES: Record<AuditAction, string> = {
 	[AUDIT_ACTIONS.FIELD_UPDATED]: '{fieldName} was updated from {oldValue} to {newValue}',
 	[AUDIT_ACTIONS.FIELD_CLEARED]: '{fieldName} ({oldValue}) was removed',
 
+	//long-text-fields
+	[AUDIT_ACTIONS.FIELD_SET_LONG]: '{fieldName} was set to {newValue}',
+	[AUDIT_ACTIONS.FIELD_UPDATED_LONG]: '{fieldName} was updated',
+	[AUDIT_ACTIONS.FIELD_CLEARED_LONG]: '{fieldName} {oldValue} was removed',
```
pinssbrett: These should not include the values in the templates. The values will be output in the details element. All three cases should work in exactly the same way - the only difference is that SET will not have a previous value and CLEARED will not have a new value.

```suggestion
	[AUDIT_ACTIONS.FIELD_SET_LONG]: '{fieldName} was set',
	[AUDIT_ACTIONS.FIELD_UPDATED_LONG]: '{fieldName} was updated',
	[AUDIT_ACTIONS.FIELD_CLEARED_LONG]: '{fieldName} was removed',
```

apps/manage/src/app/audit/actions.ts
pinssbrett: I think it's important that these action names follow the MPESC convention so that the action itself is at the end - then it's easier to read them and tell them apart:
- `LONG_FIELD_SET`
- `LONG_FIELD_UPDATED`
- `LONG_FIELD_CLEARED`

apps/manage/src/app/views/cases/case-history/view-model.ts
```diff
@@ -13,19 +13,72 @@ export interface CaseHistoryRow {
...
+				longDetails: value
+					? [
+							{
+								label: 'Show full details',
+								value
+							}
+						]
+					: undefined
+			};
+		}
```
pinssbrett: This should pass the `longDetails` for all three actions in the same way, so they all work in the same way. In the SET case the (empty) previous array element should not be passed, and in the CLEARED case the new element should not be passed (i.e. only one accordion). The label should be `Previous ${fieldName}` or `New ${fieldName}`

apps/manage/src/app/views/cases/case-history/view.njk
```diff
@@ -93,6 +83,29 @@
...
+
+							const update = function() {
+								textEl.textContent = el.open ? 'Hide full details' : 'Show full details';
+							};
+
+							update();
+							el.addEventListener('toggle', update);
+						});
+					</script>
+				{% endblock %}
```
pinssbrett: This was not asked for in the ticket - summary text should either be `Previous {fieldname}` or `New {fieldname}`. Makes it easier so you don't have to toggle the text.

pinssbrett (changes requested):
SET and CLEARED are still being handled in an overly complex way - they can all work the same way

apps/manage/src/app/audit/actions.ts
```diff
@@ -18,6 +18,11 @@ export const AUDIT_ACTIONS = {
 	FIELD_UPDATED: 'FIELD_UPDATED',
 	FIELD_CLEARED: 'FIELD_CLEARED',
 
+	// long-text fileds
+	LONG_FIELD_SET: 'FIELD_SET_LONG',
+	LONG_FIELD_UPDATED: 'FIELD_UPDATED_LONG',
+	LONG_FIELD_CLEARED: 'FIELD_CLEARED_LONG',
```
pinssbrett: These all need updating too
```suggestion
	LONG_FIELD_SET: 'LONG_FIELD_SET',
	LONG_FIELD_UPDATED: 'LONG_FIELD_UPDATED',
	LONG_FIELD_CLEARED: 'LONG_FIELD_CLEARED',
```

apps/manage/src/app/audit/actions.ts
```diff
@@ -18,6 +18,11 @@ export const AUDIT_ACTIONS = {
 	FIELD_UPDATED: 'FIELD_UPDATED',
 	FIELD_CLEARED: 'FIELD_CLEARED',
 
+	// long-text fileds
```
pinssbrett: nitpick: typo
```suggestion
	// long-text fields
```

apps/manage/src/app/audit/actions.ts
```diff
@@ -29,7 +34,7 @@ export type AuditAction = (typeof AUDIT_ACTIONS)[keyof typeof AUDIT_ACTIONS];
  * Narrows `string` to `AuditAction` when the value is a known action key.
  */
 export function isAuditAction(value: string): value is AuditAction {
-	return value in AUDIT_ACTIONS;
+	return Object.values(AUDIT_ACTIONS).includes(value as AuditAction);
```
pinssbrett: Why did this need to be changed?

pinssbrett (changes requested):
Great progress - just a final tidy up required!

pinssbrett (approved):
Great work!

PR #1152, opened 2026-07-31
feat(manage): peas 221 add view and edit case press notice tab

andrewlilley-solirius:
## Describe your changes
Addition of Press notice details tab for view and edit case section

https://pins-ds.atlassian.net/browse/PEAS-221

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/view/questions.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/questions.ts
@@ -804,6 +804,95 @@ export function getQuestions(
 			url: '',
 			editable: false
 		},
+		pressNoticeCost: {
+			type: COMPONENT_TYPES.MULTI_FIELD_INPUT, // TODO: PEAS-390 Change from multi-field to single-line once it supports prefixes
+			title: 'Press notice cost',
+			question: 'What is the cost of the press notice?',
+			fieldName: 'pressNoticeCost',
+			url: 'cost',
+			hint: 'For example, £1000.00',
+			inputFields: [
+				{
+					fieldName: 'pressNoticeCost',
+					prefix: { text: '£' },
+					formatPrefix: '£'
+				}
+			],
+			validators: [
+				new RequiredValidator('Enter the cost of the press notice'),
+				new StringValidator({
+					regex: {
+						regex: '^(?!(?:.*\\d){9,})\\d+(?:\\.\\d{1,2})?$',
+						regexMessage: 'Cost of press notice must be 8 digits or less'
+					}
+				}),
+				new StringValidator({
+					regex: {
+						regex: '^[0-9]+(\\.[0-9]{1,2})?$',
+						regexMessage: 'Cost of press notice should include numbers only'
+					}
+				})
+			],
+			viewData: {
+				extraActionButtons: [
+					{
+						text: 'Remove and save',
+						type: 'submit',
+						formaction: 'cost/remove'
+					}
+				]
+			}
+		},
+		pressNoticePlaced: {
+			type: COMPONENT_TYPES.TEXT_ENTRY,
+			title: 'Press notice placed',
+			question: 'Where has the press notice been placed?',
+			fieldName: 'pressNoticePlaced',
+			url: 'placed',
+			validators: [
+				new RequiredValidator('Enter where the press notice has been placed'),
+				new StringValidator({
+					maxLength: {
+						maxLength: 250,
+						maxLengthMessage: 'Where the press notice has been placed must be 250 characters or less'
+					}
+				})
+			],
+			viewData: {
+				extraActionButtons: [
+					{
+						text: 'Remove and save',
+						type: 'submit',
+						formaction: 'placed/remove'
+					}
+				]
+			}
+		},
+		pressNoticeReference: {
+			type: COMPONENT_TYPES.SINGLE_LINE_INPUT,
+			title: 'Press notice reference',
+			question: 'What is the reference of the press notice?',
+			fieldName: 'pressNoticeReference',
+			url: 'reference',
+			validators: [
+				new RequiredValidator('Enter press notice reference'),
+				new StringValidator({
+					maxLength: {
+						maxLength: 250,
+						maxLengthMessage: 'Press notice reference must be 250 characters or less'
+					}
+				})
+			],
+			viewData: {
+				extraActionButtons: [
+					{
+						text: 'Remove and save',
+						type: 'submit',
+						formaction: 'reference/remove'
+					}
+				]
+			}
+		},
 		pressNoticeDate: {
 			type: COMPONENT_TYPES.DATE,
 			title: 'Press notice date',
--- a/apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
@@ -31,6 +31,7 @@ import {
 import { addBusinessDays } from 'date-fns';
 import { optionalWhere } from '@pins/crowndev-lib/util/database.ts';
 import { slugify, sentenceCase } from '@pins/crowndev-lib/util/string.ts';
+import { toDecimalOrNull } from '@pins/crowndev-lib/util/numbers.ts';
 
 const DATE_FIELDS_SET = new Set<string>(S62A_DATE_FIELDS);
 const FEE_BOOLEAN_SET = new Set<string>(FEE_BOOLEAN_FIELDS);
@@ -179,6 +180,11 @@ export interface UpdateCaseAnswers {
 	wasteActivitiesDescription?: string;
 	isWasteManagementDevelopment?: YesNo;
 	manageWasteTypes?: WasteTypeItem[] | null;
+
+	//Press Notice tab
+	pressNoticeCost?: number | null;
+	pressNoticeReference?: string | null;
+	pressNoticePlaced?: string | null;
 }
 
 /**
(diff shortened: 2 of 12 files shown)


packages/database/src/schema.prisma
```diff
@@ -214,6 +214,11 @@ model S62aCase {
   eiaScreening        Boolean?
   eiaScreeningOutcome Boolean?
 
+  // Press Notice Fields
+  pressNoticeCost      Decimal?
```
paymonsattar: `pressNoticeCost` has no precision annotation, so the migration generated `DECIMAL(32,16)`. Sixteen decimal places for a currency amount is wrong and inconsistent with other fields in the schema. See cilAmount as an example.

I would suggest updating this to `pressNoticeCost Decimal? @db.Decimal(12, 2)` and regenerating the migration.

apps/manage/src/app/views/s62a/cases/view/view-model.ts
```diff
@@ -239,7 +244,7 @@ const BOOLEAN_FIELDS = Object.freeze([
  * These integer fields are in the SaveModel as strings because they are in a multi-field input.
  * TODO update multi-field input to allow number fields CROWN-1620
  */
-const INTEGER_STRING_FIELDS = Object.freeze(['siteNorthing', 'siteEasting'] as const);
+const INTEGER_STRING_FIELDS = Object.freeze(['siteNorthing', 'siteEasting', 'pressNoticeCost'] as const);
```
paymonsattar: `pressNoticeCost` is a Decimal on the DB side, but adding it to `INTEGER_STRING_FIELDS` sends it through `assignNullableDirectField`.

That means the view model receives a `Prisma.Decimal object` where the interface declares `number` - it'll likely render as `[object Object] `on the CYA rather than the cost.

Can you double check this? 

I suggest removing this from `INTEGER_STRING_FIELDS` and adding a `toNumber()` block - look at `cilAmount` as an example.

apps/manage/src/app/views/s62a/cases/view/questions.ts
```diff
@@ -796,6 +796,93 @@ export function getQuestions(
 			url: '',
 			editable: false
 		},
+		pressNoticeCost: {
+			type: COMPONENT_TYPES.MULTI_FIELD_INPUT, // TODO: update to single line input once that has been updated to have prefixes in dynamic forms
```
AlexMoulin: Can we please make a ticket to update and add the ticket number here?

apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
```diff
@@ -923,6 +929,25 @@ export class S62aCaseUpdateMapper {
...
 	}
+	/*
+	 * Maps the Press Notice tab scalar fields.
+	 */
+	private mapPressNotice(input: Prisma.S62aCaseUpdateInput): void {
+		const ans = this.answers;
+
+		if (this.hasAnswer('pressNoticeCost')) {
+			input.pressNoticeCost =
+				typeof ans.pressNoticeCost === 'number' || typeof ans.pressNoticeCost === 'string' ? ans.pressNoticeCost : null;
```
paymonsattar: The cost is passed straight through rather than converted to a `Decimal` here

For example cilAmount does:

```input.cilAmount = ans.cilAmount || ans.cilAmount === 0 ? new Prisma.Decimal(ans.cilAmount) : null;```

Could you match this? You'll also have to update the test for this as you're asserting a string there

File: apps/manage/src/app/views/s62a/cases/view/questions.ts [hunk 5]
```diff
@@ -796,6 +796,93 @@ export function getQuestions(
...
+					{
+						text: 'Remove and save',
+						type: 'submit',
+						formaction: 'cost/remove'
+					}
+				]
+			}
+		},
+		pressNoticePlaced: {
+			type: COMPONENT_TYPES.TEXT_ENTRY,
```
paymonsattar, line 840: Is there a reason you used `TEXT_AREA` here? The ticket mentions "Text input (250 characters)" and we normally use `SINGLE_LINE_INPUT` for that.

packages/database/src/seed/s62a/data-static.ts
```diff
@@ -142,7 +142,8 @@ export const VIEW_TAB_ID = Object.freeze({
 	CONTACTS: 'contacts',
 	OUTCOME: 'outcome',
 	EIA: 'eia',
-	PRE_APPLICATION: 'pre-application'
+	PRE_APPLICATION: 'pre-application',
+	PRESS: 'pressnotice'
```
paymonsattar: I think this should be `press-notice` to match the other tabs. Can you also update the section segment in `journey.ts` and the test in `journey.test.ts` to match this?

apps/manage/src/app/views/s62a/cases/view/view-model.ts
AlexMoulin: Is pressNoticeCost not a decimal?

apps/manage/src/app/views/s62a/cases/view/view-model.ts
```diff
@@ -220,6 +220,11 @@ export interface S62aCaseViewModel {
 	decisionOutcomeId?: string | null;
 	decisionDate?: Date;
 	recoveredReportSentDate?: Date;
+
```
paymonsattar, line 236: You need to add some test coverage for the new fields on this file - have a look at some of the other PRs in this epic to see examples

paymonsattar (changes requested):
Looks good but there are some things that need addressing!

apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
AlexMoulin: Does single line input return a number? And if it is a string should it parsed so it can be saved as a float?

I think the input is treated like a string.

apps/manage/src/app/views/s62a/cases/view/questions.ts
```diff
@@ -796,6 +796,93 @@ export function getQuestions(
...
+			}
+		},
+		pressNoticeReference: {
+			type: COMPONENT_TYPES.SINGLE_LINE_INPUT,
+			title: 'Press notice reference',
+			question: 'What is the reference of the press notice?',
+			fieldName: 'pressNoticeReference',
+			url: 'reference',
+			validators: [
+				new RequiredValidator('Enter press notice reference'),
```
AlexMoulin, line 871: The reference has a maximum of 250 char same as the pressNoticePlaced. Is this maximum accurate? Is they are, I wonder what the justification for user Single line input vs text entry is.

packages/database/src/schema.prisma
AlexMoulin: You probably want to set some precision on cost so that the db doesn't have more than 2 numbers after the decimal point. IIRC it's @db.Decimal(8, 2) as it has 8 numbers max

packages/database/src/schema.prisma
```diff
@@ -214,6 +214,11 @@ model S62aCase {
...
+  pressNoticePlaced    String?
+  pressNoticeReference String?
```
AlexMoulin: You max also want to reduce the maximum char for these. IIRC it defaults to nvarchar(1000) in SQL server see: https://www.prisma.io/docs/orm/v6/reference/prisma-schema-reference#string but you want a maximum of 250

AlexMoulin (changes requested):
Some comments to review, looks good overall but requires some tightening on types in mapper and db

apps/manage/src/app/views/s62a/cases/view/questions.ts
andrewlilley-solirius: Good point! Have investigated, and the larger text box for the pressNoticePlaced field is the desired UI decision to allow all text to be visible without scrolling, but also keeping the 250 max character limit. 

Given that I'll keep the TEXT_ENTRY Component there, unless there are any strong objections, and action the remaining comments

apps/manage/src/app/views/s62a/cases/view/questions.ts
andrewlilley-solirius: Good point! Have investigated, and the larger text box for the pressNoticePlaced field is the desired UI decision to allow all text to be visible without scrolling, but also keeping the 250 max character limit.   

Given that I'll keep the TEXT_ENTRY Component there, unless there are any strong objections, and action the remaining comments

apps/manage/src/app/views/s62a/cases/view/questions.ts
andrewlilley-solirius: Of course! Have created [PEAS-390](https://pins-ds.atlassian.net/browse/PEAS-390) to link to this comment for you

apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
```diff
@@ -989,6 +995,41 @@ export class S62aCaseUpdateMapper {
...
+		if (this.hasAnswer('pressNoticePlaced')) {
+			input.pressNoticePlaced = typeof ans.pressNoticePlaced === 'string' ? ans.pressNoticePlaced : null;
+		}
+
+		if (this.hasAnswer('pressNoticeReference')) {
+			input.pressNoticeReference = typeof ans.pressNoticeReference === 'string' ? ans.pressNoticeReference : null;
+		}
+	}
+
+	private toDecimalOrNull(value: unknown): Prisma.Decimal | null {
```
AlexMoulin: I think this probably needs to be moved to a util. There is a numbers util in the package that would be perfect 👍

AlexMoulin (changes requested):
Looks good, just need to move the helper to a util so we can use it elsewhere and you should be good !

AlexMoulin (dismissed):
LGTM

paymonsattar (dismissed):
Nice work, looks good

paymonsattar (approved):
LGTM

PR #1154, opened 2026-08-03
Feat/peas 217 event tab

paymonsattar:
## Describe your changes
Adds the event tab to the s62A view & edit page

https://pins-ds.atlassian.net/browse/PEAS-217

Screenshot 2026-08-03 at 11 57 18
Screenshot 2026-08-03 at 11 58 06
image

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/view/questions.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/questions.ts
@@ -21,6 +21,7 @@ import {
 	S62A_CATEGORIES,
 	S62A_PRE_APPLICATION_STATUSES,
 	S62A_STAGES,
+	SITE_VISIT_TYPES,
 	SPECIALISMS
 } from '@pins/crowndev-database/src/seed/s62a/data-static.ts';
 import {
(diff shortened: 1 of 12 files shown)


apps/manage/src/app/views/s62a/cases/view/questions.ts
```diff
@@ -1746,6 +1747,143 @@ export function getQuestions(
 					{ text: 'Remove and save', type: 'submit', formaction: 'recovered-report-sent-date/remove' }
 				]
 			}
+		},
+		noticeOfProcedureDate: {
+			type: COMPONENT_TYPES.DATE,
+			title: 'Notice of procedure date',
+			question: 'When is the Notice of procedure date?',
```
pinssbrett: Nitpick - capitalisation
```suggestion
			question: 'When is the notice of procedure date?',
```

apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
```diff
@@ -154,6 +160,19 @@ export interface UpdateCaseAnswers {
 	decisionOutcomeId?: string | null;
 	decisionDate?: Date | null;
 	recoveredReportSentDate?: Date | null;
+
+	// Event tab
+	procedureNotificationDate?: Date | null;
+	hearingDate?: Date | null;
+	prepDuration?: string | number | null;
+	sittingDuration?: string | number | null;
+	reportingDuration?: string | number | null;
```
pinssbrett: Use the actual type here for clarity. From memory this is `string` because multi-field input does not support number fields yet (TODO CROWN-1620)

```suggestion
	prepDuration?: string | null;
	sittingDuration?: string | null;
	reportingDuration?: string | null;
```

pinssbrett (changes requested):
Good, I just have a couple of small nitpicks!

apps/manage/src/app/views/s62a/cases/view/questions.ts
paymonsattar: The JIRA ticket has the capitalisation like that - I'll double check with a BA as you are right... it doesn't make sense to capitalise notice

pinssbrett (approved):
All looks good, thanks!

PR #1160, opened 2026-08-03
Feat/peas 388 adds upload and validation

oscargay:
## Describe your changes
- Adds 2 new classes: a document uploader and a validator, with tests and utils

https://pins-ds.atlassian.net/browse/PEAS-57


apps/manage/src/app/views/s62a/cases/view/folders/folder/upload/upload-documents/file-validator.ts
```diff
@@ -0,0 +1,264 @@
...
+
+			const hasEncryptedStream = container.FileIndex.some((entry) =>
+				['encryptedstream', 'encryptedpackage', 'encryptioninfo'].includes(entry.name?.toLowerCase())
+			);
+
+			if (hasEncryptedStream) return true;
+
+			const wordEntry = container.FileIndex.find((entry) => entry.name === 'WordDocument');
+			if (wordEntry && wordEntry.content && wordEntry.content.length > 0x0b) {
+				const content = wordEntry.content as number[];
```
pinssbrett: is this cast safe? the content could be a `Buffer`, `Uint8Array`, or `number[]` - would it be better to normalise it?

apps/manage/src/types/file-type.d.ts
pinssbrett, line 1: why is this needed? from what I can tell the `fileTypeFromBuffer` type defined in the package works fine

pinssbrett (changes requested):
Just a couple of small type questions

apps/manage/src/app/views/s62a/cases/view/folders/folder/upload/upload-documents/file-validator.ts
oscargay: I think these are remnants from before I correctly got CFB types working, as removing this one and the one below seems fine! Good spot, thanks

apps/manage/src/types/file-type.d.ts
oscargay: I was having a lot of issues with `file-type` not being identified in the IDE, and then some threads online saying newer versions struggle. But having said that I just deleted the file and it seems to be completely fine. Not sure what was happening but have removed! Thanks!

pinssbrett: If you do have longer-term issues with this I think it might have been due to `moduleResolution` being `node` in `jsonconfig` but `nodenext` in `tsconfig` (despite this being used in a TS file) - changing to `nodenext` may resolve it

pinssbrett (approved):
Looks good to me!

PR #1161, opened 2026-08-03
Feat/peas 389 adds routing controller view model

oscargay:
## Describe your changes
- Adds the wiring and view model frontend for the upload component
- Creates an extension of the MOJ component to make the UI more similar to prototype
- Contains 2 commits from parents that must be merged & rebased first

https://pins-ds.atlassian.net/browse/PEAS-57

## Implementation
Screenshot 2026-08-03 at 15 26 25
Screenshot 2026-08-03 at 15 26 46
Screenshot 2026-08-03 at 15 27 01


apps/manage/src/app/sass/multi-file-upload.scss
```diff
@@ -0,0 +1,47 @@
...
+		display: flex;
+		align-items: center;
+		width: 100%;
+	}
+
+	.moj-multi-file-upload__col-name {
+		flex: 0 0 65%;
+		min-width: 0;
+		word-break: break-word;
+		padding-right: 15px;
```
pinssbrett: Should we use govuk spacing scale?
```suggestion
		padding-right: govuk-spacing(3);
```

apps/manage/src/app/sass/multi-file-upload.scss
```diff
@@ -0,0 +1,47 @@
...
+
+	.moj-multi-file-upload__col-actions:empty {
+		display: none;
+	}
+
+	// This component goes Files Addes -> File Uploader by default with a black dashed dropzone
+	// But in our designs we want the reverse order with blue dashed dropzone.
+	.moj-multi-file-upload--enhanced {
+		display: flex;
+		flex-direction: column-reverse;
```
pinssbrett: Does this comply with [WCAG 1.3.2 Meaningful Sequence](https://www.w3.org/WAI/WCAG22/Understanding/meaningful-sequence.html) and [WCAG 2.4.3 Focus Order](https://www.w3.org/WAI/WCAG22/Understanding/focus-order.html)?

apps/manage/src/app/views/s62a/cases/view/folders/folder/upload/view.njk
```diff
@@ -0,0 +1,218 @@
...
+                </div>
+            </form>
+            <form id="main-submit-form" action="{{ currentUrl }}/commit" method="post">
+                <input type="hidden" name="_csrf" value="{{ _csrf }}">
+                {{ govukButton({
+                    text: "Upload"
+                }) }}
+            </form>
+
+            <script type="module" {% if cspNonce %}nonce="{{ cspNonce }}"{% endif %}>
```
pinssbrett: Up until now inline JS has been fairly light. This is getting pretty hefty - might it be time to extract client-side JS into its own files and set up a compiler so that we can get the benefits of linting, minification, etc?

apps/manage/src/app/views/s62a/cases/view/folders/folder/upload/view.njk
```diff
@@ -0,0 +1,218 @@
...
+                import { MultiFileUpload } from '/assets/js/moj-frontend.min.js';
+
+                const $mojFormContainer = document.querySelector('#upload-form-container');
+                const $mainSubmitForm = document.querySelector('#main-submit-form');
+                const $multiFileUpload = document.querySelector('[data-module="moj-multi-file-upload"]');
```
pinssbrett: is the $-prefix naming a jQuery convention? i'm not sure if it makes sense here

pinssbrett (changes requested):
i have a few comments on the frontend aspects - i've not touched on the csrf approach as i know you're discussing that separately. the client-side js thing might need to be discussed as a team (if you think it's a valid consideration)

apps/manage/src/app/sass/multi-file-upload.scss
oscargay: This is a good point to raise, I think Meaningful Sequence is probably fine but I think still good to check, but I think it would impact Focus Order definitely as it would jump back up if using a keyboard. Unfortunately, as its using a component we don't have access to the HTML, so cannot change it there... Will need to have a think about this one!

apps/manage/src/app/views/s62a/cases/view/folders/folder/upload/view.njk
oscargay: Good point, might be worth having a larger conversation about this in general with the team as you mentioned as would be good to get people's thoughts!

apps/manage/src/app/views/s62a/cases/view/folders/folder/upload/view.njk
oscargay: Ah good spot, thanks, I think I spent too many years using jquery!

apps/manage/src/app/sass/multi-file-upload.scss
oscargay: @pinssbrett - the only solution I can see for this which feels sensible, is to take the MOJ nunjucks (which is actually very simply) and just reverse them as a custom njk file and import that). I have done this and it works whilst maintaing WCAG standards?

apps/manage/src/app/views/s62a/cases/view/folders/folder/upload/view.njk
oscargay: @pinssbrett - I figured out why I was doing this on autopilot, I was looking at the MoJ component which also does the same thing, so I had to do some of them with $ because they are class variables, and then it felt better to just align them all? Not sure though

pinssbrett: ah right - i feel like aligning class variables to the MoJ component makes sense but that's a good boundary, and anything outside of that can use standard naming convention? your choice as to what you think is neatest

pinssbrett (approved):
Looks good! Thanks for rolling with the changes in direction

PR #1165, opened 2026-08-05
fix(macros): peas 371 shared list table fix to allow sort by ref not url

andrewlilley-solirius:
## Describe your changes
Fix to shared nunjucks case list table component to allow sorting by specific field. Fixes issue where table was sorting on passed url fields instead of reference strings.

Changes also made to view model, and test file, to support requested reference sorting.

Fixes failed acceptance criteria for PEAS-111: Table does not sort in descending order 

https://pins-ds.atlassian.net/browse/PEAS-371

Relates to:
https://pins-ds.atlassian.net/browse/PEAS-145

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/list/controller.ts
+++ b/apps/manage/src/app/views/s62a/cases/list/controller.ts
@@ -49,6 +49,8 @@ export function buildCaseListPage(service: ManageService): AsyncRequestHandler {
 
 		const paginationParams = createPaginationParams(req, totalItems);
 
+		console.log(s62aCasesViewModels);
+
 		return res.render('./views/s62a/cases/list/view.njk', {
 			pageTitle: 'Manage Section 62A applications',
 			s62aCasesViewModels,
--- a/apps/manage/src/app/views/s62a/cases/list/controller.ts
+++ b/apps/manage/src/app/views/s62a/cases/list/controller.ts
@@ -49,8 +49,6 @@ export function buildCaseListPage(service: ManageService): AsyncRequestHandler {
 
 		const paginationParams = createPaginationParams(req, totalItems);
 
-		console.log(s62aCasesViewModels);
-
 		return res.render('./views/s62a/cases/list/view.njk', {
 			pageTitle: 'Manage Section 62A applications',
 			s62aCasesViewModels,
--- a/apps/manage/src/app/views/s62a/cases/list/view-model.ts
+++ b/apps/manage/src/app/views/s62a/cases/list/view-model.ts
@@ -5,6 +5,7 @@ import { insertWbr, formatStatusTag } from '@pins/crowndev-lib/util/string.ts';
 
 export interface S62ACaseView {
 	id: string;
+	createdDate: Date;
 	reference: string;
 	lpaName: string | undefined;
 	applicationType?: string;
@@ -18,6 +19,7 @@ export interface S62ACaseView {
 
 export const s62aCaseSelect = {
 	id: true,
+	createdDate: true,
 	reference: true,
 	Lpa: { select: { name: true } },
 	description: true,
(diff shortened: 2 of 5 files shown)


apps/manage/src/app/views/s62a/cases/list/view.njk
```diff
@@ -46,7 +46,7 @@
 				</p>
 				{{ paginationOptions.renderPagination(paginationParams.selectedItemsPerPage, paginationParams.totalItems, baseUrl, queryParams) }}
 				<hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
-				{{ baseCaseListTable(s62aCasesViewModels,baseManageHeaders,baseManageColumns,true) }}
+				{{ baseCaseListTable(s62aCasesViewModels,baseManageHeaders,s62aManageColumns,true) }}
```
AlexMoulin, line 49: Since you are no longer using baseManageColumns, you should be able to remove it from the imports.

packages/lib/views/tables/tables.njk
```diff
@@ -80,7 +87,8 @@
 
 {% set baseManageColumns = [
 	{
-		field: "referenceLink"
+		field: "referenceLink",
+		sortField: "reference"
```
AlexMoulin, line 93: I don't think this change is required,

apps/manage/src/app/views/s62a/cases/list/controller.ts
```diff
@@ -49,6 +49,8 @@ export function buildCaseListPage(service: ManageService): AsyncRequestHandler {
 
 		const paginationParams = createPaginationParams(req, totalItems);
 
+		console.log(s62aCasesViewModels);
```
AlexMoulin: I think this console.log is a remnant of testing?

apps/manage/src/app/views/s62a/cases/list/view-model.ts
```diff
@@ -48,7 +51,8 @@ export function s62aToViewModel(s62aCases: S62ACasePayload) {
 			).map((item) => item.Organisation!.name) ?? [],
 		location: '',
 		referenceLink:
-			'<a class="govuk-link" href="/s62a/cases/' + s62aCases.id + '">' + insertWbr(s62aCases.reference) + '</a>'
+			'<a class="govuk-link" href="/s62a/cases/' + s62aCases.id + '">' + insertWbr(s62aCases.reference) + '</a>',
+		sortReference: ''
```
AlexMoulin: Might be a good to fallback on a value since your sortReference is in a if statement. You could create a function to generate the sortReference and insert it here with a nullish coalescing (??) and fallback on reference.

AlexMoulin: You may also want to consider renaming this to something a little clearer. It is a sort override for the reference column that uses reference and date: sortKey or if you want to be specific for the default on this column sortByReferenceThenDate

AlexMoulin (changes requested):
Looks good, you have got the logic down. Some minor tweaks and I think you will be good to go.

apps/manage/src/app/views/s62a/cases/list/controller.ts
andrewlilley-solirius: My apologies, tidying that up!

packages/lib/views/tables/tables.njk
andrewlilley-solirius: The only reason I would keep that is that this is used by the shared table macro in crown manage. 

For the crown back-end manage list, this would fix the current "sort by url link" bug by tying the sort to reference directly, as it was originally in the local table macros before they were abstracted. Otherwise it would sort by referenceLink if left as-is

AlexMoulin (approved):
LGTM

PR #1167, opened 2026-08-05
Feat/peas 231 add waste tab

paymonsattar:
## Describe your changes
Adds the Waste tab to the S62A case view, with a sortable table for waste types and their capacity and throughput.

- Ported TableManageListQuestion - manage list as a table
- New DefinedColumnsTableQuestion - explicit columns and cell formatters
- New MultiConditionalRadioQuestion - unit choice with per-option amount input
- New MultiConditionalNumericValidator and UniqueListFieldValidator
- Ported ManageListItemsCompleteValidator
- First use of withCondition inside a manage list
- Void capacity skipped for four waste types
- Three new lookup tables

https://pins-ds.atlassian.net/browse/PEAS-231

Screenshot 2026-08-05 at 15 24 37
Screenshot 2026-08-05 at 15 24 43
Screenshot 2026-08-05 at 15 24 50
Screenshot 2026-08-05 at 15 25 03
Screenshot 2026-08-05 at 15 25 10
Screenshot 2026-08-05 at 15 25 31

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/view/questions.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/questions.ts
@@ -22,7 +22,10 @@ import {
 	S62A_PRE_APPLICATION_STATUSES,
 	S62A_STAGES,
 	SITE_VISIT_TYPES,
-	SPECIALISMS
+	SPECIALISMS,
+	WASTE_TYPES,
+	WASTE_UNIT_ID,
+	WASTE_UNITS
 } from '@pins/crowndev-database/src/seed/s62a/data-static.ts';
 import {
 	AddressValidator,
@@ -54,6 +57,8 @@ import CILAmountLengthValidator from '@pins/crowndev-lib/forms/custom-components
 import { multiContactQuestions, createLpaContactQuestion } from '../util/question-factories.ts';
 import { getApplicantOrganisationOptions } from '../../../../views/cases/util/applicant-organisation-options.js';
 import TelephoneNumberValidator from '@pins/crowndev-lib/validators/telephone-number-validator.ts';
+import MultiConditionalNumericValidator from '@pins/crowndev-lib/forms/custom-components/multi-conditional-radio/multi-conditional-numeric-validator.ts';
+import UniqueListFieldValidator from '@pins/crowndev-lib/validators/unique-list-field-validator.ts';
 import type { EntraGroupMembers } from '#util/entra-groups.ts';
 
 type ApplicantOrg = {
@@ -107,6 +112,8 @@ export function getQuestions(
 	const hasAgent = answers?.hasAgent === BOOLEAN_OPTIONS.YES;
 
 	const applicantContactsValidator = getApplicantContactsValidator(hasAgent, isIndividual);
+
+	const throughputUnits = WASTE_UNITS.filter((u) => u.id !== WASTE_UNIT_ID.CUBIC_METRES);
 
 	const questions = {
 		reference: {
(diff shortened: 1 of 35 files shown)


apps/manage/src/app/views/s62a/cases/view/questions.ts
```diff
@@ -1884,6 +1891,149 @@ export function getQuestions(
...
+		wasteType: {
+			type: CUSTOM_COMPONENTS.MULTI_CONDITIONAL_RADIO,
+			title: 'Type of waste',
+			question: 'Which types of waste are applicable to this development?',
+			fieldName: 'wasteTypeId',
+			url: 'waste-types',
+			boldSummaryValue: true,
+			validators: [
+				new RequiredValidator('Select type of waste'),
+				new UniqueListFieldValidator({
```
oscargay, line 1969: There is a validator called `CrossQuestionValidator` in dyn forms that allows you to check answers on other fields, could that be used here to avoid having to make a new validator?

apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
```diff
@@ -517,6 +524,44 @@ export class S62aCaseUpdateMapper {
...
+
+		// Deliberately not hasAnswer(): it returns false for an empty array, so
+		// removing the last waste type would be skipped and the row would survive.
+		if (ans.manageWasteTypes === undefined) return;
+
+		const items = ans.manageWasteTypes ?? [];
+
+		input.WasteTypes = {
+			deleteMany: {},
+			create: items
```
oscargay, line 555: Just realised that inspectors does this too above, but is this deleting all `WasteTypes` for a case and re-adding them every time a user makes an update? If so, we might get issues down the line with case history because it's going to say a bunch of stuff was deleted then re-added every time a user makes a minor edit? Unless I've misunderstood this part here

packages/lib/forms/custom-components/manage-list-table/confirm.njk
```diff
@@ -0,0 +1,45 @@
+{% extends layoutTemplate %}
```
oscargay: Are you sure you need this still? Shouldn't it be identical to the other one?

packages/lib/forms/custom-components/manage-list/table/answer-summary-list.njk
oscargay, line 1: I think maybe all this stuff should go under the already existing custom dir `manage-list`, then you can reuse e.g. the confirm.njk but also it's less confusing and these can be viewed as children of the main table? Otherwise we have 2 custom directories `manage-list` and `manage-list-table` which could be confusing perhaps? What do you think

apps/manage/src/app/views/s62a/cases/view/questions.ts
```diff
@@ -1884,6 +1891,149 @@ export function getQuestions(
...
+					fieldName: unit.id,
+					label: unit.displayName,
+					classes: 'govuk-input--width-10',
+					inputmode: 'numeric'
+				}
+			})),
+			validators: [
+				new RequiredValidator('Select the unit of measurement'),
+				new ConditionalRequiredValidator('Enter the total capacity of the void'),
+				new MultiConditionalNumericValidator({
```
oscargay, line 2007: `CrossQuestionValidator` might also be able to handle this type too? Again I could just be missing something

oscargay (changes requested):
Looks good! Have a few questions, mainly around whether we can use pre-existing stuff for a few of the new things. We also have a manage list already so making sure we handle that in a way that doesn't make it confusing about which component should be used? I do think it's the right idea to keep this separated from the existing crown manage list component though... Let me know your thoughts!

apps/manage/src/app/views/s62a/cases/view/questions.ts
paymonsattar: CrossQuestionValidator compares against another field's answer, but this needs to compare against other items in the same list and exclude the item being edited - which needs req.params.manageListItemId, and the validation function signature doesn't expose req... Without that, changing anything else on an existing row would fail against its own waste type. Happy to be wrong if there's a way to reach the params I've missed.

apps/manage/src/app/views/s62a/cases/view/questions.ts
paymonsattar: It could, but it'd need one instance per unit option with the field names hard-coded, since the pairing is amount-to-its-own-unit rather than one field to one other. MultiConditionalNumericValidator derives the pairs from question.options, so adding a unit needs no change here. It's also a near-copy of ConditionalRequiredValidator, which does the same scoping for the required check. Again happy to switch if you  think it's better

apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
paymonsattar: Good catch, and you're right - the inspectors does the same. It's a wholesale replace on every save. A diff-based update (match by id, update/create/delete only what changed) would be the fix, similar to how mapApplicantOrganisations works. 

I think we should make a ticket for this to update along with inspectors when we get to case history. 

Happy to make the ticket

packages/lib/forms/custom-components/manage-list/table/answer-summary-list.njk
paymonsattar: Both good points. I'll move it all under manage-list/table/ and drop the duplicate confirm.njk - I think mine only differed by a removalPrompt fallback

packages/lib/forms/custom-components/manage-list-table/confirm.njk
paymonsattar: Have updated this!

apps/manage/src/app/views/s62a/cases/view/s62a-update-case-mapper.ts
```diff
@@ -517,6 +524,44 @@ export class S62aCaseUpdateMapper {
...
+		}
+
+		// Deliberately not hasAnswer(): it returns false for an empty array, so
+		// removing the last waste type would be skipped and the row would survive.
+		if (ans.manageWasteTypes === undefined) return;
+
+		const items = ans.manageWasteTypes ?? [];
+
+		input.WasteTypes = {
+			deleteMany: {},
```
oscargay, line 554: Based on the other conversation, shall we put a TODO with a ticket here and on inspectors to update these? Otherwise we may forget 🤔

oscargay (commented):
Looks good, just 1 comment about leaving a TODO for the `deleteMany` components? Otherwise I think all good now!

apps/manage/src/app/views/s62a/cases/view/questions.ts
```diff
@@ -1884,6 +1891,149 @@ export function getQuestions(
...
+					inputmode: 'numeric'
+				}
+			})),
+			validators: [
+				new RequiredValidator('Select the unit of measurement'),
+				new ConditionalRequiredValidator('Enter the maximum annual operational throughput'),
+				new MultiConditionalNumericValidator({
+					regexMessage: 'Maximum annual operational throughput must be a number'
+				})
+			]
```
pinssbrett, line 2042: I think these questions use custom components only to add bolding, labels and suffixes to the summary value - if `RadioQuestion` could do this, we could just use the core component. I'm going to add the ability to do this to dynamic forms. You could add a TODO here to retire the `MultiConditionalRadio` once this is implemented?

packages/lib/forms/custom-components/manage-list/table/question.ts
```diff
@@ -0,0 +1,338 @@
...
+	override addCustomDataToViewModel(viewModel: QuestionViewModel): void {
+		if (!this.section) {
+			throw new Error('Section not set for TableManageListQuestion');
+		}
+
+		super.addCustomDataToViewModel(viewModel);
+
+		this.addButtonText(viewModel);
+
+		// viewModel.question is typed `any` upstream
```
pinssbrett, line 106: Is this a bug in dynamic forms?

packages/lib/forms/custom-components/manage-list/table/question.ts
```diff
@@ -0,0 +1,338 @@
...
+	/**
+	 * For an empty list, defer to the parent with null so the correct
+	 * "not started" text and add link are shown.
+	 */
+	getAction(sectionSegment: string, journey: Journey, answer: unknown): unknown {
+		if (Array.isArray(answer) && !answer.length) {
+			return super.getAction(sectionSegment, journey, null) as unknown;
+		}
+
+		return super.getAction(sectionSegment, journey, answer) as unknown;
```
pinssbrett, line 342: There's quite a bit of type stuff in this class due to type deficiencies in dynamic-forms. Ideally this would be returning `ActionView | ActionView[] | undefined` like its parent. Can you add a TODO and a ticket to fix this file once type issues are resolved?

pinssbrett (commented):
As @oscargay has already reviewed I haven't looked at the functional stuff in detail but I have a few comments about dynamic-forms compatibility/types

apps/manage/src/app/views/s62a/cases/view/questions.ts
paymonsattar: Have added a TODO

packages/lib/forms/custom-components/manage-list/table/question.ts
paymonsattar: I wouldn't say this is a bug in dynamic forms as such... I think it's deliberate but lossy.

The type is `type QuestionViewModel = { [x: string]: any };` - I'm guessing it's been done this way so question type can put different things on the view model e.g.e label, legend, options - and rather than model each, they made it open.

pinssbrett (approved):
Thanks for adding the TODOs - all good!

oscargay (approved):
LGTM

PR #1169, opened 2026-08-05
feat: adds file table to folder page

oscargay:
## Describe your changes
- Adds a doc table to the folder page

https://pins-ds.atlassian.net/browse/PEAS-62

### DIFF
--- a/apps/manage/src/app/views/s62a/cases/view/folders/folder/view-model.test.ts
+++ b/apps/manage/src/app/views/s62a/cases/view/folders/folder/view-model.test.ts
@@ -1,0 +1,85 @@
+import { describe, it } from 'node:test';
+import assert from 'node:assert';
+import { createDocumentsViewModel, type DocumentWithFolder } from './view-model.ts';
+
+describe('createDocumentsViewModel', () => {
+	const mockPreviewMimeTypes = ['application/pdf', 'image/jpeg'];
+
+	const createMockDoc = (overrides: Partial<DocumentWithFolder> = {}): DocumentWithFolder => {
+		return {
+			id: 'doc-123',
+			fileName: 'site-plan.pdf',
+			size: BigInt(1048576),
+			mimeType: 'application/pdf',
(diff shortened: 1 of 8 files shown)


apps/manage/src/app/views/s62a/cases/view/folders/folder/view-model.ts
```diff
@@ -0,0 +1,80 @@
...
+): DocumentViewModel[] {
+	return documents.map((doc) => {
+		const dateObj = new Date(doc.uploadedDate);
+		const sizeNum = Number(doc.size);
+
+		const downloadHref = `/`; // TODO PEAS-58: update to real download link in next ticket
+		const deleteHref = `/`; // TODO PEAS-293: update to real delete link in next ticket
+
+		return {
+			asdfL: 'asdf',
```
paymonsattar: Is this a debug or something left over?

apps/manage/src/app/views/s62a/cases/view/folders/folder/view-model.test.ts
```diff
@@ -0,0 +1,85 @@
+import { describe, it } from 'node:test';
+import assert from 'node:assert';
+import { createDocumentsViewModel, type DocumentWithFolder } from './view-model.ts';
+
+describe('createDocumentsViewModel', () => {
+	const mockPreviewMimeTypes = ['application/pdf', 'image/jpeg'] as any;
```
paymonsattar: I don't think you need the `as any` cast here. Works without it

paymonsattar (changes requested):
Two small comments, otherwise looks good!

apps/manage/src/app/views/s62a/cases/view/folders/folder/view-model.ts
oscargay: 🤦 - I was testing some type issue and clearly forgot about it

paymonsattar (approved):

PR #1173, opened 2026-08-06
feat: adds document downloading

oscargay:
## Describe your changes
- Adds document downloading to S62A

https://pins-ds.atlassian.net/browse/PEAS-58


packages/lib/util/file.ts
```diff
@@ -0,0 +1,39 @@
...
+export function isolateFileNameFromExtension(fileName: string) {
+	const dotIndex = fileName.lastIndexOf('.');
+
+	if (dotIndex === -1) return [fileName, ''];
+
+	const baseName = fileName.substring(0, dotIndex);
+	const extension = fileName.substring(dotIndex);
+
+	return [baseName, extension];
+}
```
pinssbrett: Could use `import { parse } from 'node:path';` here for simplicity?

apps/manage/src/app/views/s62a/cases/view/folders/folder/controller.ts
```diff
@@ -112,7 +116,21 @@ export function buildViewCaseFolder(service: ManageService): AsyncRequestHandler
...
+ * Reads session data and clears it to avoid it being displayed > once.
+ */
+function readAndClearSessionData(req: Request) {
+	const id = getStringParam(req.params, 'id');
+
+	const errorSummary = readSessionData(req, id, 'filesErrors', false, 'folder');
+	clearSessionData(req, id, 'filesErrors', 'folder');
+
+	return [errorSummary];
+}
```
pinssbrett: this feels like a familiar pattern in the codebase. could it be its own helper in `util/session.ts`?

pinssbrett (changes requested):
Looks good! A couple of minor nitpicks

oscargay:
@pinssbrett - back with you, thanks!

I addressed both comments. Created 2 new utils, one for the read and wipe functionality as you mentioned and also a uri validator that just makes sure any redirect uri we build from anything passed by the user is safe

packages/lib/util/uri.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Verify that a given URI is valid for a redirect
+ * This should only allow relative paths.
+ */
+export function isValidRedirectUri(uri: unknown) {
```
pinssbrett, line 5: Does this need reconciling with the function of the same name in `apps/manage/src/app/auth/controller.js`?

pinssbrett (changes requested):
Looks good, but I think there is a duplication to resolve with the redirect checker

oscargay:
@pinssbrett - thanks, back with you, have updated to use the shared component. Also noticed some styling got lost in a recent rebase onto main, so updated a class name too

pinssbrett (approved):

PR #1177, opened 2026-08-07
fix(manage): record audit entry with 'Unknown-user' when userId is missing

RobBroadhead30:
## Describe your changes
This pull request improves the audit logging behavior in the case update workflow to ensure that audit entries are always recorded, even when a user ID is missing. Previously, audit entries would be skipped if the user ID was not present, but now a placeholder value is used and a warning is logged. The changes also add a new test to confirm this behavior.

**Audit logging improvements:**

* Changed the audit logging logic in `update-case.ts` to use `'Unknown-user'` as the `userId` when it is missing, ensuring audit entries are always recorded and logging a warning instead of skipping the entry. [[1]](diffhunk://#diff-6e26e0918f33a6815dfe31b2f9f6330d01a493c8d7442ae5515860b08f50f381R473-R475) [[2]](diffhunk://#diff-6e26e0918f33a6815dfe31b2f9f6330d01a493c8d7442ae5515860b08f50f381L507-R507)

**Testing enhancements:**

* Added a test in `update-case.test.ts` to verify that an audit entry is recorded with `userId: 'Unknown-user'` when the user ID is not present.

https://pins-ds.atlassian.net/browse/CROWN-1705
image

### DIFF
--- a/apps/manage/src/app/views/cases/view/update-case.test.ts
+++ b/apps/manage/src/app/views/cases/view/update-case.test.ts
@@ -2737,7 +2737,7 @@ describe('audit recording', () => {
 		assert.strictEqual(mockAudit.recordMany.mock.callCount(), 0);
 	});
 
-	it('should skip audit and log warning when userId is undefined', async () => {
+	it('should record audit entry when userId is not present', async () => {
 		const logger = mockLogger();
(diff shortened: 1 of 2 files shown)


paymonsattar (approved):
LGTM

